PR #68, opened 2024-02-21
Add initial ONNX test suite with import, compile, run scripts.

ScottTodd:
Progress on https://github.com/openxla/iree/issues/16372

_Meta: this includes over 5000 generated files totaling 12MB. I'm not sure yet how much we want checked in, in which repository, and with what folder structure. Most of the diffs are in the last commit, for which it might be easier to review by looking at a few examples like this: https://github.com/ScottTodd/SHARK-TestSuite/tree/onnx-testsuite/iree_tests/onnx/node/generated/test_abs or this: https://github.com/ScottTodd/SHARK-TestSuite/tree/onnx-testsuite/iree_tests/onnx/node/generated/test_matmul_2d_

This has a few components to seed a new type of test suite for IREE:

* `iree_tests/onnx/import_tests.py` to import upstream ONNX tests into a converted directory structure that is compatible with `iree-compile` and `iree-run-module`
* `iree_tests/compile_tests.py` to compile converted tests following that structure to `.vmfb` files
* `iree_tests/run_tests.py` to run compiled `.vmfb` files using test inputs and expected outputs

A few things fall out from this setup naturally:

* We can add new test suites or cases, including from other frameworks (PJRT/StableHLO, TFLite/TOSA, etc.), with similar generic `import_*` scripts or bespoke scripts for individual models of interest
* We can fan out to multiple compilation and run configurations
* We can run the entire suite or choose subsets
* Since the suite uses release artifacts (`iree-import-onnx` as a preprocessing step then just `iree-compile` and `iree-run-module`), it can run on nightly releases, on presubmit artifacts, or on developer builds. It can also run using out of tree plugins and other downstream packages

This still has several rough edges. The largest missing pieces I can see now are:

* Integration into GitHub Actions. We could start with an action in this repository or jump straight to an action in IREE that clones this repository and runs the tests with artifacts from a release or [pre,post]-submit run.
* Wiring up to a test runner like pytest or ctest. The .py scripts currently run serially and don't report results or dump reproducer files. We will also want to parameterize each test case with passing/XFAIL configurations.


iree_tests/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Available test suites
+
+### ONNX test suite
+
+The ONNX test suite is a conversion of the upstream test suite from
+[onnx/`onnx/backend/test/`](../third_party/onnx/onnx/backend/test/):
+
+* Python sources in [onnx/`onnx/backend/test/case/`](../third_party/onnx/onnx/backend/test/case)
+* Generated `.onnx` and `[input,output]_[0-9]+.pb` files in [onnx/`onnx/backend/test/data/`](../third_party/onnx/onnx/backend/test/data)
```
ScottTodd, line 104: Aw.. GitHub doesn't follow these links into submodules (VSCode just follows the local paths). Might replace with links to the web: https://github.com/onnx/onnx/tree/main/onnx/backend/test/

iree_tests/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.vmfb
+# TODO(scotttodd): compile into a build dir instead of the source dir
+config_*.txt
```
ScottTodd, line 3: If using pytest, this can be replaced with https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html
If ctest, could use the build directory

For local iteration I do like keeping all related files in one folder though (could have pytest/ctest copy into the build/temp folder)

iree_tests/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+Developer quickstart:
+
+```bash
+python -m pip install iree-compiler iree-runtime
+python -m pip install -r ./iree_tests/onnx/requirements.txt
+
+python ./iree_tests/onnx/import_tests.py
+python ./iree_tests/compile_tests.py
+python ./iree_tests/run_tests.py
```
ScottTodd, line 40: At least the compile and run scripts will likely be replaced with pytest files / plugins.

* https://docs.pytest.org/en/stable/
* https://docs.pytest.org/en/stable/how-to/writing_plugins.html#conftest-py-plugins
* https://docs.pytest.org/en/stable/example/nonpython.html#yaml-plugin

iree_tests/compile_tests.py
```diff
@@ -0,0 +1,105 @@
...
+THIS_DIR = Path(__file__).parent
+REPO_ROOT = THIS_DIR.parent
+
+# Test directory generated by iree_tests/onnx/scripts/generate_onnx_tests.py
+# TODO(scotttodd): flag to run across any compatible test suite folder
+TEST_SOURCE_ROOT = REPO_ROOT / "iree_tests/onnx/node/generated"
+
+# Write lists of tests that passed/failed to compile.
+COMPILE_SUCCESSES_FILE = REPO_ROOT / "iree_tests/onnx/node/compile_successes.txt"
+COMPILE_FAILURES_FILE = REPO_ROOT / "iree_tests/onnx/node/compile_failures.txt"
```
ScottTodd, line 20: This will turn into some sort of configuration file in the test runner, starting in this repo and then moving into downstream projects. IREE can run the tests it wants with its own out-of-band labels for which test cases pass/fail, iree-amd-aie can do the same, etc.

iree_tests/README.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+Testing follows several stages:
+
+1. Import
+2. Compile
+3. Run
```
ScottTodd, line 47: We'll also want a 'benchmark' stage for some programs / configurations. I'm not sure yet if that should chain off of the same test runner / framework, or just leverage the raw test suite files. Tightly coupling running benchmarks to building and testing in IREE's CI has been painful in some ways.

iree_tests/compile_tests.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # TODO(scotttodd): clear .vmfb files from source/build dir?
+
+    print(f"Compiling tests in '{TEST_SOURCE_ROOT}'")
+
+    print("******************************************************************")
+    passed_compiles = []
+    failed_compiles = []
+    # TODO(scotttodd): parallelize this (or move into a test runner like pytest)
+    for i in range(len(test_dir_paths)):
```
ScottTodd, line 80: https://pypi.org/project/pytest-xdist/

ScottTodd (commented):
Chatted about this a bit with @stellaraccident offline. Summarized some discussion points in comments.

I figure we'd land this and iterate, but could also keep the work on a branch/fork for some time. Landing on the main branch in a shared repo helps with visibility and ensures we're on a path to using the code across projects.

iree_tests/README.md
ScottTodd: Here's a GitHub Actions workflow to run the 'compile' and 'run' scripts: https://github.com/ScottTodd/SHARK-TestSuite/blob/test-githubaction/.github/workflows/test_iree.yml

Sample runs:
* https://github.com/ScottTodd/SHARK-TestSuite/actions/runs/8024507737/job/21923232663 (all tests, 'run' is hanging for some reason on the Linux server...)
* https://github.com/ScottTodd/SHARK-TestSuite/actions/runs/8024846159/job/21924226218 (limited to 20 tests)

I'm going to set up pytest next so we can get better test output, define run configurations, check XFAIL/XPASS, etc.

ScottTodd: and conftest.py that replaces the `compile_tests.py` script (`run_tests.py` port coming soon too): https://github.com/ScottTodd/SHARK-TestSuite/blob/iree-pytest/iree_tests/onnx/basic/conftest.py

posted a more detailed update on how that fits together here: https://github.com/openxla/iree/issues/16372#issuecomment-1962143140

kumardeepakamd:
I am thinking whether to have these tests run additionally from e2eshark/run.py using -testsfile list where list is list of tests and a switch to have an integrated report run. @saienduri is adding a --mode  turbine  option to e2eshark run.py to be able to run any pytorch test with and without dtype bf16. It would give ways to test different code path like through ort ep, converting to bf16 etc. using same test in addition to integrated reporting.

kumardeepakamd:
Pretty good initiative. I like it a lot. I am thinking how to make it very useful for torch MLIR development flow as well. Some thoughts are as below., We can iterate on it:
1) Since we want this approach to scale, I am thinking not checking in the generated file but giving ability to generate the test and test output of a particular kind: onnx, torch mlir or linalg may work better. Similar approach can be applied to tests gotten from other frameworks, this also keeps the tests current. The IREE developer may want to start from torch MLIR, but a front end developer adding support for ONNX OP in torch MLIR or ONNX importer will like to have onnx test. Also, an end to end onnx test that compares the run result in ONNX runtime vs iree-run-module will be very useful. 
2) If the ability to install tests in a particular directory is provided, then e2eshark run.py could take an option to invoke installation as onnx tests with likely with a a little tweak to generated files, and then run these tests e2e and generated compbined report. That will allow flushing out many issues in front end
Let us chat this week more on this.

stellaraccident:
At least for iree, we want test input artifacts snapshotted somewhere vs generated on every run. The CI volume and variants that need to check this stuff is too great to not have that step.

But that can be a bespoke place that gets updated with "cooked" ready to test inputs materialized.

Would be nice if this was not completely different tooling flows like it is today (versus more like priming the cache and then saving it -- same flow just starting in the middle for backend CI).

iree_tests/README.md
ScottTodd, line 47: <sub>(commenting here for threading)</sub>

@kumardeepakamd : 

> 1. Since we want this approach to scale, I am thinking not checking in the generated file but giving ability to generate the test and test output of a particular kind: onnx, torch mlir or linalg may work better. Similar approach can be applied to tests gotten from other frameworks, this also keeps the tests current. The IREE developer may want to start from torch MLIR, but a front end developer adding support for ONNX OP in torch MLIR or ONNX importer will like to have onnx test. Also, an end to end onnx test that compares the run result in ONNX runtime vs iree-run-module will be very useful.

---

@stellaraccident :

> At least for iree, we want test input artifacts snapshotted somewhere vs generated on every run. The CI volume and variants that need to check this stuff is too great to not have that step.
>
> But that can be a bespoke place that gets updated with "cooked" ready to test inputs materialized.
>
> Would be nice if this was not completely different tooling flows like it is today (versus more like priming the cache and then saving it -- same flow just starting in the middle for backend CI).

---

+1 to Stella's comments. For CI and downstream development flows, we need to decouple from the frontend framework. The full e2e test flows are useful for a certain segment of developers and users, but inserting a snapshot layer like this will keep complexity down.

I'm imagining that
* IREE will consume input MLIR, inputs, and outputs - _not_ ONNX models or Python sources. IREE will use a pinned version of the test suite repo that may lag behind the latest imported files.
  * This is currently just using checked in .mlir files, but we could also extend to download .mlir files from a cloud bucket if they grow beyond a certain size (beyond that git LFS can handle). The ONNX test suite has some code for "real" models generating inputs/outputs from a reference backend in Python. I want to avoid putting logic in the test suite for as long as possible.
  * For large tests, we can use IREE parameters too. Splats would save on space but we'd need to be careful about expected outputs / program behavior without real weights.
* The test repo itself could test the ONNX -> MLIR importer and update the generated files when appropriate. Similar story for JAX/StableHLO, TFLite/TOSA, and other programs in the test suite.

saienduri:
This is a really cool setup! 
Here are my thoughts for overall initial CI:
I was thinking that we can leverage this setup for IREE CI (maybe add an actions workflow in IREE that clones this subfolder and runs the tests on different backends with the torch MLIR as Scott was saying). Then, we will have e2e CI in this repo. That CI will comprise of a job that runs the Test Suite on all the models through different modes (onnx, turbine, direct) and dtypes (with and without bf16) and generates report artifact that are hosted on GitHub Pages which can be mailed to our team. We can keep track of all the models status (pinpoint where exactly they fail) and model regressions everyday. This will also let us have a page where anyone can see reports of our status for all models.  We will also have a Turbine Tank CI here. I'm going to move turbine tank implementation from Turbine repo to here. That will run all the models e2e through Turbine and upload the inline and exported param versions of the mlir to Azure. If e2e test fails, it will be marked as such so that it won't be sent out to IREE CI. The final step will be to connect the turbine tank to Scott's setup here. We will essentially add the mlir generated by tank into this setup everyday (create script that downloads the most recent torch mlir blobs that passed e2e from Azure for every model we want tested in IREE and run this script in another nightly actions job). This torch mlir can be run in iree against supported backend configurations and tested on every PR created there. I think this way we have all our testing and CI centralized here. This also allows all our testing to be connected and up to date as changes happen at every level.

ScottTodd:
I have changes stacked on top of this at https://github.com/ScottTodd/SHARK-TestSuite/pull/1. How would we like to proceed to getting these merged?

Note that the `iree_tests/compile_tests.py` and `iree_tests/run_tests.py` scripts are now completely replaced by a [pytest](https://docs.pytest.org/en/stable/) `iree_tests/conftest.py` script.

* I could merge that work into this branch and we could fully review before landing
* (I think my preference) we could land this as-is and then I could send a second PR with the improvements

iree_tests/README.md
```diff
@@ -0,0 +1,111 @@
...
+  inputs and outputs for one test case
+* `test_data_flags.txt` is a flagfile for use with
+  `iree-run-module --flagfile=test_data_flags.txt` of the format:
+
+  ```
+  --input=@input_0.npy
+  --expected_output=@output_0.npy
+  ```
+
+## Running the test suite
```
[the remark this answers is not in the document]
kumardeepakamd: Okay, that is fair. sounds good.

iree_tests/README.md
kumardeepakamd: Checking in .mlir for iree make sense. I guess, as needed that can be refreshed.

kumardeepakamd:
Yeah, it will help to land this PR and then iterate over it and have e2eshark, iree_tests and Sai's tests be there side by side and leverage components and evolve it. I can land it if there is not objection.

Issue #46, opened 2024-02-12
Phase names and log names inconsistent


renxida:
The name of the phases should be consistent with the names of the log files.

renxida:
e.g. phase onnx-import has its log file called torch-onnx.log.

kumardeepakamd:
https://github.com/nod-ai/SHARK-TestSuite/pull/73 
Good observation, thanks for reporting, uses current phase name now.

kumardeepakamd:
Fixed it. Please check.

PR #73, opened 2024-02-26
Fix issue #46 and make post process pipeline use common code

kumardeepakamd:
https://github.com/nod-ai/SHARK-TestSuite/issues/46 fixed, uses now current phase name
The tools.stubs.commonutils is included in run.py and model.py to allow building any arbitrary post processing pipeline to allow meaningful comparison of outputs for quantized models

### DIFF
--- a/e2eshark/run.py
+++ b/e2eshark/run.py
@@ -10,6 +10,11 @@ import argparse
 import numpy as np
 import torch, io
 import struct, pickle, tabulate, statistics
+from pathlib import Path
+
+# Need to allow invocation of run.py from anywhere
+sys.path.append(Path(__file__).parent)
+from tools.stubs.commonutils import applyPostProcessPipeline
 
 # Put full path to your Torch MLIR ( https://github.com/llvm/torch-mlir ) build
 # This can be overwritten by command line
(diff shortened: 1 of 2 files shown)


e2eshark/tools/stubs/commonutils.py
```diff
@@ -43,22 +43,31 @@ def flatten_tuples(tup):
     return test_op_list
 
 
+# listOfFunctionArgRetTuple = (function, [args other than input], isReturnTuple, indexOfTupleForTupleReturn)
+def applyPostProcessPipeline(item, functionPipeLine):
```
ScottTodd, line 47: Comment doesn't match the function parameter name? Should this be 'functionPipeLine'?

You could also use a named tuple or class instead of an unstructured tuple: https://docs.python.org/3/library/collections.html#collections.namedtuple

e2eshark/run.py
```diff
@@ -73,13 +78,13 @@ def concatenateFiles(inpfile1, inpfile2, outfile):
     ofile.close()
 
 
-def loadTestCheckedDictionary():
-    testCheckedDict = None
+def loadE2ESHARK_CHECK_DICT():
+    e2esharkDict = None
```
ScottTodd, line 82: function name isn't following typical naming style - fix?

kumardeepakamd:
Thanks for catching them, fixed, merging now.

PR #75, opened 2024-02-27
Setup pytest and GHA workflow for IREE test suite.

ScottTodd:
Progress on https://github.com/openxla/iree/issues/16372

This replaces the `compile_tests.py` and `run_tests.py` scripts with a [pytest](https://docs.pytest.org/en/stable/) `conftest.py` script. The README contains instructions for running with pytest and debugging the pytest configuration itself.

Users of this test suite can define config files and pass them to the pytest script via the `IREE_TEST_CONFIG_FILES` environment variable. Each config file sets flags to use as well as a list of XFAIL tests. We can expand that with more options as this matures, but I tried to keep it fairly simple to start.

* Sample run of the new workflow on my fork: https://github.com/ScottTodd/SHARK-TestSuite/actions/runs/8057581337/job/22008943380
* Sample output with error messages for one compiler and one runtime failure: https://gist.github.com/ScottTodd/5ad27d31f8f3f30a590a86bb97a0885b
* Sample output with all runtime failures (by dropping all runtime XFAILs from the config): https://gist.github.com/ScottTodd/540dde8cd67c1820d73b4ff697c806ed

## Future work

* Add workflows to the IREE repo along with forked versions of the config files
* Add new tests to the test suite (full torch models, stablehlo programs, etc.)
  * May need to adjust test case naming and discovery as part of this (e.g. test names in the config files are not currently scoped to a particular folder / suite)
* Explore test report generation options
* Consider switching `import_tests.py` to pytest as well, with some automation in this repo for catching regressions and picking up improvements in `iree-import-onnx`


stellaraccident (approved):
This looks like great progress!

kumardeepakamd (approved):
Can go ahead and land it and then iterate. Getting better with each commit!

PR #76, opened 2024-02-27
add turbine mode + configurations and support for all models (image classification, t5 seq2seq, language)

saienduri:
This commit adds the implementation of turbine mode in the test suite which plugs into the classical flow with the turbinemodel stub so we can leverage the aot export. It also adds configuration and support for all the models we want to track. I tested every model in turbine and direct mode, and it works as I'd expect it to. One thing to note, the extra tensor args popped up again with the latest torch-mlir. It seems to be a different issue than before from a quick look at the mlir, and I can tackle it once initial CI is up and running. Two models fail at model run in turbine mode due to turbine still being at torch 2.1, which will hopefully be resolved soon.


e2eshark/pytorch/models/bart-large/model.py
```diff
@@ -0,0 +1,45 @@
...
+model_response = model.generate(
+    E2ESHARK_CHECK["input"],
+    do_sample=True,
+    top_k=50,
+    max_length=100,
+    top_p=0.95,
+    temperature=1.0,
+)
+print("Prompt:", prompt)
+print("Response:", tokenizer.decode(model_response[0]))
```
ScottTodd, line 40: What sorts of responses are these models returning? Are they a few lines of text?

e2eshark/pytorch/models/vicuna-13b-v1.3/model.py
```diff
@@ -0,0 +1,45 @@
...
+# import from e2eshark/tools to allow running in current dir, for run through
+# run.pl, commutils is symbolically linked to allow any rundir to work
+sys.path.insert(0, "../../../tools/stubs")
+from commonutils import E2ESHARK_CHECK_DEF
+
+# Create an instance of it for this test
+E2ESHARK_CHECK = dict(E2ESHARK_CHECK_DEF)
+
+
+test_modelname = "lmsys/vicuna-13b-v1.3"
```
ScottTodd, line 16: For each of these files, I'd personally find a README or at least a comment pointing at the model origin useful. WDYT?

e.g. https://huggingface.co/lmsys/vicuna-13b-v1.3 for this

Could also add an index page in the `pytorch/models` directory root with the list of supported models (possibly with a table tracking which platforms / configurations are known to be working, or with performance results -- depends on what this repo will be responsible and if that information is also available elsewhere)

e2eshark/pytorch/models/whisper-base/model.py
```diff
@@ -0,0 +1,45 @@
...
+E2ESHARK_CHECK["input"] = encoding["input_ids"].cpu()
+E2ESHARK_CHECK["output"] = model(E2ESHARK_CHECK["input"])
+model_response = model.generate(
+    E2ESHARK_CHECK["input"],
+    do_sample=True,
+    top_k=50,
+    max_length=100,
+    top_p=0.95,
+    temperature=1.0,
+)
```
ScottTodd, line 38: Are any of these parameters useful to tune in these tests? Would a developer want a way to set them across all models in the repo (e.g. with a CLI flag or environment variable)?

e2eshark/run.py
```diff
@@ -331,8 +331,13 @@ def runCodeGeneration(
         print(f"Test {testName} failed [{curphase}]")
         return 1
     logfilename = curphase + ".log"
+    # use same compile command we do in turbine
     commandname = (
-        "/tools/iree-compile --iree-input-demote-i64-to-i32 --iree-hal-target-backends="
+        "/tools/iree-compile --iree-input-type=torch --mlir-print-debuginfo "
+        + "--mlir-print-op-on-diagnostic=false --iree-llvmcpu-target-cpu-features=host "
+        + "--iree-llvmcpu-target-triple=x86_64-linux-gnu --iree-stream-resource-index-bits=64 "
```
ScottTodd, line 338: Would prefer to not hardcode this script to only run on Linux :)

In the IREE repo we also have benchmarks running with both default flags and with 'experimental' flags that are still being evaluated. Could do something similar here. Tests should use default flags wherever possible.

e2eshark/run.py
```diff
@@ -1116,7 +1096,7 @@ def main():
         "--mode",
         choices=["direct", "turbine", "onnx", "ort", "vaiml"],
         default="onnx",
-        help="direct=Fx/TS->torch-mlir, turbine=integrated-e2e, onnx=exportonnx-to-torch-mlir, ort=exportonnx-to-ortep",
+        help="direct=Fx/TS->torch-mlir, turbine=aot-export->torch-mlir, onnx=exportonnx-to-torch-mlir, ort=exportonnx-to-ortep",
```
ScottTodd, line 1099: Can you update https://github.com/nod-ai/SHARK-TestSuite/blob/main/e2eshark/README.md to talk about how to use `--mode=turbine`? (Also please feel free to restructure the file as needed, these sorts of docs get much easier to read through the more people try to follow them and propose improvements)

e2eshark/tools/stubs/turbinemodel.py
```diff
@@ -0,0 +1,112 @@
...
+msg = "The script to run a model test"
+parser = argparse.ArgumentParser(description=msg, epilog="")
+
+parser.add_argument(
+    "-d",
+    "--todtype",
+    choices=["default", "fp32", "fp16", "bf16"],
+    default="default",
+    help="If not default, casts model and input to given data type if framework supports model.to(dtype) and tensor.to(dtype)",
+)
```
ScottTodd, line 26: Does `model.to(dtype)` change the entire model...? How does that even work? (Lots of "real" workloads are expected to only reduce precision in a few specific places)

e2eshark/tools/stubs/turbinemodel.py
```diff
@@ -0,0 +1,112 @@
...
+    "--outfileprefix",
+    default="model",
+    help="Prefix of output files written by this model",
+)
+parser.add_argument(
+    "-m",
+    "--mode",
+    choices=["direct", "onnx", "ort", "turbine"],
+    default="direct",
+    help="Generate torch MLIR, ONNX or ONNX plus ONNX RT stub",
```
ScottTodd, line 38: I haven't read through enough of this setup to understand this. Should `turbinemodel.py` also reference the other `--mode` options...?

kumardeepakamd: For turbine, will we expect onnx, ort? Maybe not. If turbine is the only mode, then maybe can get rid of --mode and not send it at all. Less code to deal with and clearer.

kumardeepakamd (commented):
Thank you for doing this. This is great effort. Brings so many good models in action to be able to test PyTorch direct, ONNX and Turbine input flows.

e2eshark/pytorch/models/bart-large/model.py
```diff
@@ -0,0 +1,45 @@
+import sys, argparse
+import torch
```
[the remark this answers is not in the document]
kumardeepakamd: If code was not copied from elsewhere -- the standard way to use a pretrained model is fine -- then should go ahead and add copyright straight

e2eshark/tools/stubs/turbinemodel.py
saienduri: Yeah, turbine is the only possible mode that hits the turbinemodel.py stub file. I'll go ahead and drop it then.

e2eshark/run.py
saienduri: Yup, will edit the README with instructions on how to use turbine mode.

e2eshark/run.py
saienduri: Ok, will revert back to original options. Don't think they're necessary, just wanted consistency between turbine and here. But, default flags should do the trick. (I don't think we should be messing too much with the backend configurations here anyways)

e2eshark/pytorch/models/bart-large/model.py
saienduri: Yeah, just a few lines of text

e2eshark/pytorch/models/vicuna-13b-v1.3/model.py
saienduri: These status reports are something that we are planning to generate daily through CI, but I'll add a list of the passing models for each mode.

e2eshark/pytorch/models/whisper-base/model.py
saienduri: It's just to give out some example output to the user. It isn't used for anything on developer/testing end.

kumardeepakamd (approved):
Looks great. Will land this to start using it. Send a separate PR for README.md update with some good examples and details.

Issue #49, opened 2024-02-13
Scripts should support installed packages, not just build directories


ScottTodd:
Code like this should be generalized: https://github.com/nod-ai/SHARK-TestSuite/blob/a2aef2b1ca7293600464ab32624fbc4aac717279/e2eshark/run.py#L425-L428 https://github.com/nod-ai/SHARK-TestSuite/blob/a2aef2b1ca7293600464ab32624fbc4aac717279/e2eshark/run.py#L844-L848

Properly installed `iree-compiler` and `iree-runtime` packages from pip can just put console scripts on `$PATH`. Checking `$PATH` first would be a good default, then configuration via flags or environment variables could override if needed.

For developers iterating on tests, the tests could be exported out of Python, the scripts could point at locally built tools, or developers can build Python packages and install them locally (e.g. with https://github.com/openxla/iree/tree/main/build_tools/python_deploy).

ScottTodd:
I'm also not sure why a torch_mlir build is needed... are there equivalent Python APIs or console scripts there? For this sort of test suite to be used on CI it should operate on release artifacts directly.

ScottTodd:
I still see a requirement to build torch-mlir from source. That should be replaced with a pip-installable package (turbine?)

stellaraccident:
There shouldn't be any run/test time dependency on a torch mlir python package.

I've only looked briefly at the above, but I think there is a missing helper class... Something like `ToolHelper` for invoking the various importers. There should be a switch or env var there to control which tools get used if there is a desire to support raw torch mlir dev tooling.

ScottTodd:
Yeah, I'm spinning here trying to get a compatible torch-mlir build on my Windows machine (such that `import torch_mlir` works). I mainly want to export .mlir files (and ideally input/output .npy files) from https://github.com/nod-ai/SHARK-TestSuite/tree/main/e2eshark/pytorch/models, but needing to balance local builds juuuust right is a time sink.

ScottTodd:
Similarly, these deps should be included directly in requirements files, not sourced from another repo:

https://github.com/nod-ai/SHARK-TestSuite/blob/5caa805f5b729e66414546253795b39ca9c14f32/e2eshark/README.md?plain=1#L125-L132

stellaraccident:
[...]

PR #81, opened 2024-03-01
IREE build is made optional, fix models to allow onnx export, handle multiple input onnx export

kumardeepakamd:


e2eshark/run.py
```diff
@@ -1157,9 +1148,13 @@ def main():
         help="Please select a dir with large free space to cache all torch, hf, turbine_tank model data",
         required=True,
     )
-    
+
     args = parser.parse_args()
     cache_dir = args.cachedir
+    if not os.path.exists(cache_dir):
```
saienduri, line 1154: I have this in other PR. Maybe remove this so we have everything related to cache_dir documented there?

saienduri (approved):
Looks good to me! Nice that user has option to skip the whole iree build now.

e2eshark/run.py
```diff
@@ -333,8 +324,12 @@ def runCodeGeneration(
         print(f"Test {testName} failed [{curphase}]")
         return 1
     logfilename = curphase + ".log"
-    commandname = (
-        "/tools/iree-compile --iree-input-demote-i64-to-i32 --iree-hal-target-backends="
+    commandname = ""
+    if SHARED_IREE_BUILD:
+        commandname = SHARED_IREE_BUILD + "/tools/"
```
ScottTodd, line 329: Binary (non-Python) distributions are also useful to support. Those don't include a `tools` folder (they are flatter)

e.g.
```
iree-dist/
  bin/
    iree-compile
    iree-run-module
  include/
    iree/...
      (header files)
  lib/
    libIREECompiler.so
```

The most generic way to support both is to just point at the binary directory, either `bin/` from an install/dist like that or `tools/` from a build directory

ScottTodd (commented):
Is https://github.com/nod-ai/SHARK-TestSuite/issues/49 fixed now?

PR #82, opened 2024-03-01
Add directory + space checks for cache

saienduri:
This commit adds checks to see if the directory exists. If it does, it also checks that there is a reasonable amount of space to run models. If not, it will throw a warning and let the user know that they should pick a directory with more space available if intention is to run models.


kumardeepakamd (approved):
Looks good.

PR #99, opened 2024-03-07
add cleanup + uploading features

saienduri:
This commit allows the user to specify cleanup and upload command line options. The cleanup option allows the user to run the tests with space efficiency (removing mlir and vmfb files by default). Which types of files are removed can be changed by the user. Also, there is an uploading feature introduced. This allows the user to specify a list of tests in the `gold/upload_list.txt` file. If the test that the user is running falls in this list, the file types that the user choses are uploaded to an Azure space. They are organized here with unique identifier based on test name + date + UTC time so we won't try uploading to the same name. Also, after running the suite, a `upload_urls.txt` file is created. In this file, you can find a mapping of the test name to Azure link. This way the user doesn't have to deal with Azure. More details can be found in code.

### DIFF
--- a/e2eshark/run.py
+++ b/e2eshark/run.py
@@ -16,11 +16,11 @@ import warnings
 import datetime
 from azure.storage.blob import BlobServiceClient
 import json
+from multiprocessing import Manager
 
 # Need to allow invocation of run.py from anywhere
 sys.path.append(Path(__file__).parent)
 from tools.stubs.commonutils import applyPostProcessPipeline
-from multiprocessing import Manager
 
 def getTestsList(frameworkname, test_types):
     testsList = []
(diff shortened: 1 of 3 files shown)


e2eshark/run.py
```diff
@@ -1117,6 +1305,16 @@ def main():
...
+    parser.add_argument(
+        "--upload",
+        help="Upload to azure storage based on uploadtestsfile",
+        action="store_true",
+        default=False,
+    )
+    parser.add_argument(
+        "--uploadtestsfile",
+        help="A file with lists of tests that should be uploaded",
+    )
```
kumardeepakamd, line 1317: Looks like --uploadtestsfile is needed everytime --upload is used? If so, it would make sense to just have switch. You can get rid of --upload switch. If --uploadtestsfile is used, you have upload on.

kumardeepakamd: Also, please indicate to user where the files have been backed up and how to get it.

kumardeepakamd (commented):
Good to see cleanup and upload having independent behavior.

e2eshark/run.py
saienduri: Makes sense, did both

kumardeepakamd (approved):

PR #104, opened 2024-03-11
Initial CI [WIP]

saienduri:


.github/workflows/test_e2eshark.yml
```diff
@@ -0,0 +1,175 @@
...
+      
+      - name: "Setting up Python"
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{matrix.version}}
+
+      - name: "Setup Python venv"
+        run: python -m venv ${IREE_VENV_DIR}
+
+      - name: Build iree
```
ScottTodd, line 110: This (and all the other build steps) should be replaced with a pip install. No need to checkout a repo with all the submodules, configure toolchains (CMake, ninja, etc.), and spent CI time rebuilding from source.

.github/workflows/test_e2eshark.yml
```diff
@@ -0,0 +1,175 @@
...
+          source ${IREE_VENV_DIR}/bin/activate
+          git submodule update --init
+          pip install --upgrade pip
+          pip install -r ../torch-mlir/requirements.txt
+          pip install -r ../torch-mlir/torchvision-requirements.txt
+          pip install ../torch-mlir/torch-mlir-wheel/torch_mlir-0.0.1-cp311-cp311-linux_x86_64.whl
+          pip install -r ./e2eshark/requirements.txt
+          pip install -f https://openxla.github.io/iree/pip-release-links.html --upgrade -r ../turbine/core/iree-requirements.txt
+          pip install -e ../turbine/core[testing]
+          pip install -e ../turbine/models
```
ScottTodd, line 174: Yikes that's a lot of pip installs. We should be able to get that down to a much shorter list... (turbine, iree, local requirements for this repo)

.github/workflows/test_e2eshark.yml
```diff
@@ -0,0 +1,175 @@
...
+            -DCMAKE_C_COMPILER=clang \
+            -DCMAKE_CXX_COMPILER=clang++ \
+            -DIREE_ENABLE_LLD=ON
+          cmake --build ../iree-build/
+        working-directory: ./iree
+  
+  turbine:
+    strategy:
+      matrix:
+        os: [nodai-amdgpu-w7900-x86-64]
```
ScottTodd, line 130: nit: this is a machine type, not an OS?

.github/workflows/test_e2eshark.yml
saienduri: Yeah, it was a mess getting the whole source build into CI. Should be able to just use the pip install.

.github/workflows/test_e2eshark.yml
saienduri: Yeah for space reasons to run all the models, I was thinking we use this machine

.github/workflows/test_e2eshark.yml
```diff
@@ -0,0 +1,175 @@
+# Copyright 2024 Advanced Micro Devices
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: E2ESHARK Test Suite
+on:
+  pull_request:
```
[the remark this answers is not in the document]
saienduri: Nice! Will do that

PR #107, opened 2024-03-12
Adding tolerance parameters --atol and --rtol to run.py script

afalkenberg1:
Parameters added to provide tolerances for comparing the test results.
--atol
--rtol 

For example: 
python ./run.py -c /home/ubuntu/E2E_MLIR_IREE/TORCH_MLIR/torch-mlir_main_af/build -i /home/ubuntu/E2E_MLIR_IREE/IREE/iree-build --report --cachedir /home/ubuntu/E2E_MLIR_IREE/CacheDir --tests pytorch/models/opt-1.3b --atol 1e-03 --rtol 1e-03

### DIFF
--- a/e2eshark/run.py
+++ b/e2eshark/run.py
@@ -465,11 +465,11 @@ def runCodeGeneration(
 #    fp32: atol=1e-05, rtol=1.3e-06
 def getTolerances(args, torchdtype):
     if args.atol != False and args.rtol != False:
-        return (float(args.atol) , float(args.rtol))
+        return (float(args.atol), float(args.rtol))
     elif args.atol != False:
-        return (float(args.atol) , 1e-03)
+        return (float(args.atol), 1e-03)
     elif args.rtol != False:
-        return (1e-03, float(args.rtol))    
+        return (1e-03, float(args.rtol))
     elif torchdtype == torch.bfloat16:
         return (1e-02, 1e-01)
     elif torchdtype == torch.float16:
--- a/e2eshark/run.py
+++ b/e2eshark/run.py
@@ -463,14 +463,19 @@ def runCodeGeneration(
 #    bf16: atol=1e-05, rtol=1.6e-02
 #    fp16: atol=1e-05, rtol=1e-03
 #    fp32: atol=1e-05, rtol=1.3e-06
-def getTolerances(torchdtype):
-    if torchdtype == torch.bfloat16:
+def getTolerances(args, torchdtype):
+    if args.atol != False and args.rtol != False:
+        return (float(args.atol) , float(args.rtol))
+    elif args.atol != False:
+        return (float(args.atol) , 1e-03)
+    elif args.rtol != False:
+        return (1e-03, float(args.rtol))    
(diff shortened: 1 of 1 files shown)


e2eshark/run.py
```diff
@@ -463,15 +463,20 @@ def runCodeGeneration(
 #    bf16: atol=1e-05, rtol=1.6e-02
 #    fp16: atol=1e-05, rtol=1e-03
 #    fp32: atol=1e-05, rtol=1.3e-06
-def getTolerances(torchdtype):
-    if torchdtype == torch.bfloat16:
+def getTolerances(args, torchdtype):
```
kumardeepakamd, line 466: @afalkenberg1 thank you for adding this option. I think we should make it override default regardless of the dtype. This is a usewr option to override the default. So, if rtol and atol has been passed, then use it whatever it is.

e2eshark/run.py
```diff
@@ -1344,6 +1349,16 @@ def main():
         action="store_true",
         default=False,
     )
+    parser.add_argument(
```
kumardeepakamd, line 1352: Let us be consistent with types. Since they are floating point numbers, can we simply require it to take a floating point value with no default? Then you check if args.rtol then use it, else use the default set in getTolerances function.
For each of atol and rtol:

parser.add_argument(
        "--atol",
        help="Set atol tolerance value",
        type=float,
    )

parser.add_argument(
        "--rtol",
        help="Set rtol tolerance value",
        type=float,
    )

e2eshark/run.py
afalkenberg1: ok I am dropping the case

e2eshark/run.py
afalkenberg1: ok changed it ; will push in couple of mins

e2eshark/run.py
```diff
@@ -459,19 +459,17 @@ def runCodeGeneration(
...
+# Pytorch torch.testing:
+def getTolerances(args):
+    if args.atol and args.rtol:
+        return (float(args.atol), float(args.rtol))
+    elif args.atol:
+        return (float(args.atol), 1e-03)
+    elif args.rtol:
+        return (1e-03, float(args.rtol))
     else:
         return (1e-04, 1e-04)
```
ScottTodd, line 471: If these values aren't conditional on anything at runtime (like the data type or specific test case), you can just set default values on the argument parser instead of using a helper function.

```python
    parser.add_argument(
        "--rtol",
        help="Set rtol tolerance parameter",
        type=float,
        default=1e03,
    )
    parser.add_argument(
        "--atol",
        help="Set atol tolerance parameter",
        type=float,
        default=1e03,
    )

    inferencematched = torch.allclose(infooutput, goldoutput, args.rtol, args.atol)
```

ScottTodd (approved):

e2eshark/run.py
afalkenberg1: yes actually I did not even think about that ; thanks now it is much cleaner

PR #108, opened 2024-03-13
add output selection and postprocessing for more accurate validation

saienduri:


kumardeepakamd (commented):
Looks good to me except one question: Can we not simply use 'postprocess' pipeline to get 'output_for_validation'?( torch.index_select, [0], False, 0) or (torch.clone, [], True [0]) or a custom function added in tools.stub.commonutils or something else like that could be used to get desired index.

saienduri:
Yeah true. We can do that, but regardless of post processing or not, we should always only be comparing the logits or last hidden states. The rest of the outputs are not important to us and not used for validation in general (past key values, other hidden states, attentions). That is why I chose this way, so regardless of whether the postprocessing flag is set, the right outputs are always compared for validation.

e2eshark/pytorch/models/bart-large/model.py
```diff
@@ -35,6 +35,7 @@
 encoding = tokenizer(prompt, return_tensors="pt")
 E2ESHARK_CHECK["input"] = encoding["input_ids"].cpu()
 E2ESHARK_CHECK["output"] = model(E2ESHARK_CHECK["input"])
+E2ESHARK_CHECK["output_for_validation"] = [E2ESHARK_CHECK["output"][0]]
```
ScottTodd, line 38: > There were two main problems. We were comparing every output generated by the model, when we should just be looking at the important outputs (ones that influence our prediction).

I'm trying to understand how this look outside of Python (specifically for test cases like those in `iree_tests`).

Does the program (.mlir file) still return multiple outputs, but now the `run.py` test runner here only checks that one output matches the reference? Should the other outputs be allowed to differ? Those seem important to test too, as they are part of the program definition.

I'm skeptical of this if my understanding is correct. I might prefer to change the model itself to only return outputs that are useful instead of teach the test runner to ignore outputs.

We can technically get the same behavior in `iree-run-module` (which `iree_tests` uses) with "(ignored") mixed in with other checks, as in:
```
--expected_output=f32=-2
--expected_output="(ignored)"
--expected_output=@output_2.npy
--expected_output=f32=@output_3.bin
```

saienduri: Yeah, so from how we do validation in other shark infra and just in general, we only look at the logits. This is what is used to predict the next token and it's ok if there are differences in other differences as long as the logits match in terms of accuracy validation. Yes, you are correct. The program (.mlir file) still return multiple outputs, but now the run.py test runner here only checks certain outputs for validation. We can't change the model itslef as it is pulled from huggingface and the other outputs are still used in future forward methods. It just isn't important in validation (okay if it differs a bit) as long as the logits match. Oh nice, the iree-run-module utility is pretty cool. Maybe we can switch to that, but current setup works too.

ScottTodd: Okay cool. So when we import test cases from this into iree_tests (or elsewhere), we'll want to remember which output index to test with which values, and which to ignore.

For example, this https://github.com/ScottTodd/SHARK-TestSuite/blob/9104b9b3b535b45c40597cfd7ae33a3d0b39b795/iree_tests/pytorch/models/opt-125M/real_weights_data_flags.txt#L1-L6 could just check the first value and use `(ignored)` for the others

ScottTodd (approved):

e2eshark/pytorch/models/bart-large/model.py
saienduri: I see, that sounds good. (so far it's just been the first index for all models)

PR #110, opened 2024-03-15
take atol and rtol as pair like torch.testing requires and have good known defaults

kumardeepakamd:
Add ERROR prefix for error messages

### DIFF
--- a/e2eshark/run.py
+++ b/e2eshark/run.py
@@ -23,6 +23,7 @@ from multiprocessing import Manager
 sys.path.append(Path(__file__).parent)
 from tools.stubs.commonutils import applyPostProcessPipeline
 
+
 def getTestsList(frameworkname, test_types):
     testsList = []
     for test_type in test_types:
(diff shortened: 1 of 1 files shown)


e2eshark/run.py
```diff
@@ -458,10 +470,22 @@ def runCodeGeneration(
...
+    atol, rtol = (1e-05, 1.3e-06)
+    if args.tolerance:
+        atol, rtol = tuple(args.tolerance)
+    elif torchdtype == torch.bfloat16:
+        atol, rtol = (1e-02, 1e-01)
+    elif torchdtype == torch.float16:
+        atol, rtol = (1e-04, 1e-03)
+    else:
+        atol, rtol = (1e-04, 1e-04)
+    return (atol, rtol)
```
ScottTodd, line 487: can simplify control flow here
```suggestion
    if args.tolerance:
        return tuple(args.tolerance)
    if torchdtype == torch.bfloat16:
        return (1e-02, 1e-01)
    if torchdtype == torch.float16:
        return (1e-04, 1e-03)
    return (1e-04, 1e-04)
```

ScottTodd (approved):

saienduri (approved):
LGTM!

e2eshark/run.py
kumardeepakamd: Sorry had missed that. Will do in next PR.

PR #115, opened 2024-03-19
Added a -torchtotosa parameter

afalkenberg1:
The underlying torch-mlir also has a torch-tosa implementation, which has a pretty good coverage. 
I needed this for some experimental results, so I implemented it here. 
Nevertheless IREE probably is not supporting it so I only used it to look at the tosa.mlir results, whereas the rest of the 
pipeline failed. If it is of interest it could be landed.  Thanks


kumardeepakamd:
I talked to Andreas. We agreed to not take it up as we want end to end inference runnable tests in this suite and our focus is torch mlir -- linalg path. Will close it.

PR #141, opened 2024-03-29
Added testcase for SiLU

afalkenberg1:
I added a testcase for SiLU as of Elliott's request.


afalkenberg1:
SiLU is a Op which might be used in certain environments and it is resolved into sigmoid and mul.

kumardeepakamd (approved):

PR #142, opened 2024-03-29
Add onnx.LSTM operator test

renxida:


renxida:
Note that this is very basic and does not test some of the optional inputs / attributes of onnx.LSTM

renxida:
Also note that it only runs with `--torchtolinalg`

zjgarvey (approved):
Nice annotations on inputs/ outputs

PR #146, opened 2024-04-02
Update Output Comparison for pytest

PhaneeshB:
- This PR changes the result comparison method from iree-run-module comparison to numpy based comparison of iree_outputs and golden_outputs

Progress on https://github.com/openxla/iree/issues/16674


.gitignore
```diff
@@ -28,6 +28,7 @@
 *.safetensors
 *.gguf
 *.vmfb
+iree_tests/onnx/node/generated/test_*/iree_output_*.npy
```
ScottTodd, line 31: The `iree_tests/` folder has its own `.gitignore` file to keep the top level file cleaner: https://github.com/nod-ai/SHARK-TestSuite/blob/main/iree_tests/.gitignore

iree_tests/conftest.py
```diff
@@ -327,8 +332,23 @@ def test_compile(self):
...
+    def test_numerical_accuracy(self):
+        num_iree_output_files = len(list(self.test_cwd.glob("iree_output_*.npy")))
+        num_output_files = len(list(self.test_cwd.glob("output_*.npy")))
+        if num_iree_output_files != num_output_files:
+            raise AssertionError(f"Number of golden outputs ({num_output_files}) and iree outputs ({num_iree_output_files}) dont match")
+
+        for i in range(num_output_files):
+            iree_output = load((self.test_cwd / f"iree_output_{i}.npy"))
+            golden_output = load((self.test_cwd / f"output_{i}.npy"))
+            assert_allclose(iree_output, golden_output, atol=self.atol, rtol=self.rtol, equal_nan=False)
```
ScottTodd, line 351: I'd like to continue using `--expected_output`, at least until we can prove that comparison in C++ is no longer sufficient.

* We have data types that numpy does not support. For that we can use binary files instead of .npy

This style of testing aims to have as thin a test runner as possible, leaning mostly on the native tools themselves. Right now all the test runner does is
1. discover test cases
2. run native tools (`iree-compile`, `iree-run-module`) with flags
3. check the return codes

By having a thin test runner with a narrow set of responsibilities, other test runner implementations are possible and results are easier to reproduce outside of the test environment.

* Someone could write a ctest (or Bazel, or something else) test runner that uses these tools. Those runners may not have as direct access to Python utilities like numpy. We should also be able to run tests on systems without Python (Android, the web, etc.)
* Test case reproducers are just commands to run. This changes that to be more complicated - "to reproduce, run `iree-run-module ...` then run this numpy code"

---

How about we first see if we can modify https://github.com/openxla/iree/blob/main/runtime/src/iree/tooling/comparison.cc to be more permissive with numpy data type mismatches, or switch the expected outputs from numpy to binary files?

iree_tests/conftest.py
```diff
@@ -293,6 +295,9 @@ def __init__(self, spec, **kwargs):
         self.run_args.extend(self.spec.iree_run_module_flags)
         self.run_args.append(f"--flagfile={self.spec.data_flagfile_name}")
 
+        self.atol=1e-05
+        self.rtol=1e-06
```
ScottTodd, line 299: We can put comparison thresholds in the flagfiles themselves, rather than make that a property of the test runner

ScottTodd (changes requested):

PhaneeshB:
Alternate approach to keep using expected output added here https://github.com/nod-ai/SHARK-TestSuite/pull/212

ScottTodd:
Closing in favor of https://github.com/nod-ai/SHARK-TestSuite/pull/212

PR #152, opened 2024-04-03
Support for sdxl pipeline (testing)

saienduri:
This commit adds support for testing all sdxl submodels


iree_tests/pytorch/models/sdxl-prompt-encoder-tank/model.mlirbc
ScottTodd, line 1: Please use Git LFS for these files. See https://github.com/nod-ai/SHARK-TestSuite/pull/147 , I think you just need to run `git lfs install` and then git will complain and convert files for you (might depend on your client, git version, etc.)

iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/real_weights_data_flags.txt
```diff
@@ -0,0 +1,8 @@
+--parameters=model=real_weights.irpa
+--module=sdxl_scheduled_unet_pipeline_fp16_.vmfb
+--input=1x4x128x128xf16=@inference_input.0.bin
+--input=2x64x2048xf16=@inference_input.1.bin
+--input=2x1280xf16=@inference_input.2.bin
+--input=1xf16=@inference_input.3.bin
+--expected_output=1x4x128x128xf16=@inference_output.0.bin
+--expected_f16_threshold=1.5f
```
ScottTodd, line 8: Is `1.5f` a large or small threshold for f16? I'm not really sure with these programs...

iree_tests/pytorch/models/sdxl-prompt-encoder-tank/test_cases.json
ScottTodd, line 1: Please document how to generate these `-tank` files / test cases.

Documentation go around https://github.com/nod-ai/SHARK-TestSuite/tree/main/iree_tests#pytorch-models or we can add a README file in one of the subdirectories.

iree_tests/configs/config_sdxl_cpu_llvm_task.json
```diff
@@ -0,0 +1,17 @@
...
+      "--iree-llvmcpu-distribution-size=32"
+    ],
+    "iree_run_module_flags": [
+      "--device=local-task"
+    ],
+    "skip_compile_tests": ["opt-125M", "resnet50", "sdxl-vae-decode-tank", "sdxl-prompt-encoder-tank"],
+    "skip_run_tests": ["opt-125M", "resnet50", "sdxl-vae-decode-tank", "sdxl-prompt-encoder-tank"],
+    "expected_compile_failures": [],
+    "expected_run_failures": []
+  }
```
ScottTodd, line 17: nit: A few of these files are missing a final newline character.

(If a file does not end in a newline character in Unix, it is not technically a file: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file. All tools I've seen are robust to that, but it's still the spec)

You can set your editor to automatically insert final newline characters (and also trim excess ones). For example, in VSCode: 
image

iree_tests/conftest.py
```diff
@@ -328,6 +338,13 @@ def test_run(self):
         proc = subprocess.run(self.run_args, capture_output=True, cwd=self.test_cwd)
         if proc.returncode != 0:
             raise IreeRunException(proc, self.test_cwd, self.compile_args)
+        
+    def test_benchmark(self):
+        proc = subprocess.run(self.benchmark_args, capture_output=True, cwd=self.test_cwd)
+        if proc.returncode != 0:
+            raise IreeRunException(proc, self.test_cwd, self.compile_args)
+        outs = proc.stdout.decode("utf-8")
+        print(f"Stdout benchmark:\n{outs}\n")
```
ScottTodd, line 347: Before this sort of change lands, let's think a bit about what we actually want coverage for. I'm skeptical about having benchmarks built into the same testing flow... though the suite that Stella set up has these:
* https://github.com/openxla/iree/blob/2c88e49d184b621c763a5cfb4af693f8dcbc6a07/experimental/regression_suite/ireers/fixtures.py#L81-L94
* https://github.com/openxla/iree/blob/2c88e49d184b621c763a5cfb4af693f8dcbc6a07/experimental/regression_suite/tests/pregenerated/test_llama2.py#L128-L148

iree_tests/pytorch/models/sdxl-prompt-encoder-tank/model.mlirbc
ScottTodd: image

Nice!

saienduri: Yeah just figured it out, thanks for the info!

iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/real_weights_data_flags.txt
saienduri: Dropped it down to 0.8, still pretty high. There are some cpu numerics that Ean is looking into. Once he figures it out and merges the sdxl branch in turbine, we can update any flags if needed and tighten the accuracy threshold.

iree_tests/configs/config_sdxl_cpu_llvm_task.json
saienduri: Cool, I've been manually checking and adding before merging a PR. This will help :). Thanks for the tip.

iree_tests/conftest.py
saienduri: Just because we care so much about sdxl perf, I think it would be great to have it included in this flow. I didn't look into adding a whole separate flow for it or making it very scalable because I doubt we will be adding benchmarks for anything else. Probably just hardcoded commands for the backends we want to test perf on. I was thinking we can evaluate and iterate if this becomes a bigger utility but for now, just went for easiest/simple add for benchmarking. Here is an example log with everything running: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8543035287/job/23405926540

ScottTodd: The value of continuous benchmarks is clear, but I want to be careful about how we integrate them. For this PR, can you leave the benchmarks off and focus on just adding the sdxl models? A follow-up PR can then add benchmarking.

I'd at least like to take some time to work through the specific requirements before jumping straight to an implementation. For example:
* What metrics/artifacts do we want from benchmarking?
  * Each model in isolation? Full pipeline latency? Just dispatch time?
* What do we want done with benchmark results / artifacts?
  * The in-tree benchmarks in IREE submit results to a dashboard (that should use a queryable database...), upload Tracy files to cloud storage, and comment on pending pull requests with results summaries
* Where do we want benchmarks to run?
  * Right after tests, on presubmit to IREE?
  * In a separate job, on separate runners?

I'm also wondering if we want to use pytest as the benchmark runner (either with the existing conftest.py or a forked one), or if we would want to use another runner (could start with a pile of scripts, just using the same test suite source files)

ScottTodd: It might be reasonable to start with `pytest -k benchmark` or `pytest iree_tests/benchmarks` that just run `iree-benchmark-module` instead of `iree-run-module` and then let developers dig through the GitHub Actions logs to see results, but I'm worried about going down the path of building an entirely new benchmark "framework" when we already have https://github.com/openxla/iree-comparative-benchmark and https://github.com/openxla/iree/tree/main/build_tools/benchmarks (building something new is likely going to make sense, at least in the short term, but this stuff gets complicated very quickly)

ScottTodd: > this stuff gets complicated very quickly

For example: https://github.com/openxla/community/blob/main/rfcs/20230505-benchmarking-strategy.md

saienduri: Yeah, I definitely understand where you're coming from. I also thought it would be cool to have all the sdxl stuff in one place (model accuracy and benchmarking results) for developers to easily see results. At least for now, I am treating this more of a corner case. I'm all onboard to change it down the line if we feel the need to.

ScottTodd: At a minimum I'd put the benchmark changes in their own pull request. Then I'd also have review comments like:

* `.github/workflows/test_iree.yml` could be forked to `.github/workflows/benchmark_iree.yml` (particularly for the `nodai-amdgpu-w7900-x86-64` matrix runner config change)
* `iree_tests/conftest.py` should not hardcode any paths. We can add new config (JSON, txt, etc.) files as needed and have a programmatic branch on whether to construct a "test" test case and/or a "benchmark" test case

saienduri: > > this stuff gets complicated very quickly
> 
> For example: https://github.com/openxla/community/blob/main/rfcs/20230505-benchmarking-strategy.md

Wow, this is a very methodical, detailed way to approach it :0

saienduri: > At a minimum I'd put the benchmark changes in their own pull request. Then I'd also have review comments like:
> 
> * `.github/workflows/test_iree.yml` could be forked to `.github/workflows/benchmark_iree.yml` (particularly for the `nodai-amdgpu-w7900-x86-64` matrix runner config change)
> * `iree_tests/conftest.py` should not hardcode any paths. We can add new config (JSON, txt, etc.) files as needed and have a programmatic branch on whether to construct a "test" test case and/or a "benchmark" test case

Makes sense, will create another PR for it

saienduri: Also the runner change is not just for benchmarking. Just one of the sdxl submodels takes a hr to run on the github runners we were using before. It was this exact job that made me make the switch lol: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8505671886

ScottTodd: Ah... I'd rather not switch to the GPU machine because CPU tests are slow. That might be what we have as runners today, but we should be a bit more principled about it.

We can split the tests across runners:
* ONNX tests on a standard GitHub-hosted CPU machine
* Full model tests on a large self-hosted CPU (or GPU) machine, possibly split even further by model

saienduri: Ah yeah, that's a good idea. I will also make ONNX testing and model testing different jobs then (think only way to split by runner is having them on different jobs).

ScottTodd: > (think only way to split by runner is having them on different jobs).

I believe that can be done with a matrix too: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-using-a-multi-dimension-matrix

Could then check other matrix parameters to choose which steps to run... maybe like this:

```yml
jobs:
  iree_tests:
    strategy:
      matrix:
        configuration:
          - runner: ubuntu-22.04
            test_command: "pytest iree_tests/simple --durations=0"
          - runner: ubuntu-22.04
            test_command: "pytest iree_tests/onnx/node/generated -n auto -rpfE --timeout=30 --retries 2 --retry-delay 5 --durations=10"
    runs-on: ${{ matrix.runner }}
    steps:
      ...
```

I'm also referencing https://github.com/openxla/iree/blob/573ff1ff02347266ed747dd316cefaeb4c710396/.github/workflows/ci.yml#L749-L784 (probably tons of other files to reference across GitHub, but that's what I know already...)

ScottTodd: If we wanted to plug in to the in-tree benchmark infrastructure that IREE has, we'd want a PR like https://github.com/openxla/iree/pull/16965 . That would feed into https://perf.iree.dev/ and PR comments, but it doesn't also test correctness out of the box, can be tricky to update (multiple Python files, coupled with GitHub Actions), and puts input files / parameters behind a few levels of abstraction that make it harder to run locally.

saienduri: Yeah, I also wonder how that would work because we need to essentially compile multiple submodels and then use their vmfbs for the pipeline. Not sure if it is setup for a pipeline structure.

saienduri: Also, I've placed the onnx and model tests in different jobs. I think that's best for this suite. Because they don't depend on each other and are running independently on different machines, I don't think we need the sequential steps. This way we have parallel execution which can help for scalability in future. Once we get more machines, splitting on models also would be great :)

saienduri: Here is the PR for benchmarking that should be landed after this one: https://github.com/nod-ai/SHARK-TestSuite/pull/155. Feel free to add on notes there for future reference

.github/workflows/test_iree.yml
```diff
@@ -67,12 +67,49 @@ jobs:
...
+      - name: "Installing other Python requirements"
+        run: |
+          source ${VENV_DIR}/bin/activate
+          python3 -m pip install -r iree_tests/requirements.txt
       # TODO(scotttodd): add a local cache for these large files to a persistent runner
       - name: "Downloading remote files for real weight model tests"
         run: |
           source ${VENV_DIR}/bin/activate
-          python3 iree_tests/download_remote_files.py
+          python3 iree_tests/download_remote_files.py --root-dir pytorch/models
```
ScottTodd, line 109: Oof this is taking 12 minutes to download: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8546974154/job/23418304591?pr=152#step:7:14

We'll really want that to be cached for presubmit.

.github/workflows/test_iree.yml
```diff
@@ -67,12 +67,49 @@ jobs:
           source ${VENV_DIR}/bin/activate
           pytest iree_tests/onnx/node/generated -n auto -rpfE --timeout=30 --retries 2 --retry-delay 5 --durations=10
 
+  linux_x86_64_w7900_gpu_models:
+    strategy:
+      matrix:
+        os: [nodai-amdgpu-w7900-x86-64]
+    name: Linux (x86_64 w7900) Models GPU
+    runs-on: ${{matrix.os}}
```
ScottTodd, line 75: Are you planning on expanding this matrix? As this is now, it's just making the names verbose and adding complexity to the yaml file.

* Matrix jobs are initially hidden under the group: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8546974154?pr=152 image
* Long 'check' name: image

.github/workflows/test_iree.yml
```diff
@@ -67,12 +67,49 @@ jobs:
           source ${VENV_DIR}/bin/activate
           pytest iree_tests/onnx/node/generated -n auto -rpfE --timeout=30 --retries 2 --retry-delay 5 --durations=10
 
+  linux_x86_64_w7900_gpu_models:
+    strategy:
+      matrix:
+        os: [nodai-amdgpu-w7900-x86-64]
```
ScottTodd, line 73: If you keep the matrix, change "os" to something that matches the text and how it is used (maybe "runner"?)

.github/workflows/test_iree.yml
```diff
@@ -67,12 +67,49 @@ jobs:
...
           source ${VENV_DIR}/bin/activate
-          python3 iree_tests/download_remote_files.py
+          python3 iree_tests/download_remote_files.py --root-dir pytorch/models
       - name: "Running real weight model tests"
+        env:
+            IREE_TEST_CONFIG_FILES: iree_tests/configs/config_sdxl_cpu_llvm_task.json
         run: |
           source ${VENV_DIR}/bin/activate
-          pytest iree_tests -n auto -k real_weights -rpfE --timeout=600 --retries 2 --retry-delay 5 --durations=0
+          pytest iree_tests/pytorch/models -s -n auto -k real_weights -rpfE --timeout=1200 --retries 2 --retry-delay 5 --durations=0
```
ScottTodd, line 115: The config file should match the tests that are running.

Either
* Rename the config file to `config_pytorch_models_cpu_llvm_task.json`
* Change the test to only run SDXL tests, e.g. `pytest -k sdxl` (possibly `pytest -k sdxl -k real_weights` ... not sure about the syntax for multiple filters)

iree_tests/configs/config_sdxl_cpu_llvm_task.json
```diff
@@ -0,0 +1,17 @@
+{
+    "config_name": "sdxl_cpu_llvm_task",
+    "iree_compile_flags" : [
+      "--iree-hal-target-backends=llvm-cpu",
+      "--iree-llvmcpu-target-triple=x86_64-linux-gnu",
```
ScottTodd, line 5: Since this is running on the same machine that is compiling, the default of `llvm::sys::getProcessTriple()` should be fine. Prefer to not hardcode a test file to a specific platform (Linux here), since developers may want to run on other platforms (Windows in my case)

iree_tests/configs/config_sdxl_cpu_llvm_task.json
```diff
@@ -0,0 +1,17 @@
...
+      "--iree-llvmcpu-target-cpu-features=host",
+      "--iree-llvmcpu-fail-on-out-of-bounds-stack-allocation=false",
+      "--iree-llvmcpu-distribution-size=32"
```
ScottTodd, line 8: Are these flags _needed_ for these models? Prefer to use default flags for tests.

iree_tests/pytorch/models/sdxl-prompt-encoder-tank/test_cases.json
```diff
@@ -0,0 +1,29 @@
...
+            "inference_input.3.bin",
+            "inference_output.0.bin",
+            "inference_output.1.bin",
+            "real_weights.irpa"
+          ]
+        }
+      ]
+    }
+  ]
+}
```
ScottTodd, line 29: Seeing more of these files, I'm still thinking about how to keep them easy to update. Might refactor to separate JSON files like

test_case_splats.json:
```json
{
  "name": "splats",
  "runtime_flagfile": "splat_data_flags.txt",
  "remote_files": []
}
```

test_case_real_weights.json:
```json
{
  "name": "real_weights",
  "runtime_flagfile": "real_weights_data_flags.txt",
  "remote_files": [
    "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sdxl-prompt-encoder/inference_input.0.bin",
    "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sdxl-prompt-encoder/inference_input.1.bin",
    "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sdxl-prompt-encoder/inference_input.2.bin",
    "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sdxl-prompt-encoder/inference_input.3.bin",
    "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sdxl-prompt-encoder/inference_output.0.bin",
    "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sdxl-prompt-encoder/inference_output.1.bin",
    "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sdxl-prompt-encoder/real_weights.irpa"
  ]
}
```

.github/workflows/test_iree.yml
ScottTodd, line 115: We probably don't want `-n auto`. That's using 64 workers: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8546974154/job/23418304591?pr=152#step:8:23

It miiiight work out, but multiple large program _compiles_ will compete for the CPU and multiple large program _runs_ will compete for the GPU. I guess ideally the compiler could be using the CPU while the runtime is using the GPU. We also need to be very careful with that for benchmarks - we don't want one benchmark influencing another.

One thing we can do is have multi-stage builds where a CPU machine runs the compiler and a GPU machine runs the tests/benchmarks. IREE's in-tree benchmarks follow that model. As long as the large model weights are externalized into parameter files and the .vmfb files are small, we can limit where we upload/download large files to just where they are needed. If we want to get really fancy, we could spin up GPU machines at the same time as CPU machines and have the GPU machines start to fetch weights (orrr just have them cached already) while the CPU machines work on compiling.

ScottTodd (changes requested):

.github/workflows/test_iree.yml
saienduri: Yeah, I removed `-n auto` on the benchmarking part for that reason. But, sure, we can remove it here also and figure out how we want to proceed in the future. A pipeline on the runner side of things would be cool. The cpu machines compile and pass the vmfbs to the gpu runners to run, but also might just be making things complicated. I'm thinking in terms of how long the CI run will take, instead of 64, do you think lowering it into 4 or something would work? (or is that also too much competition)

iree_tests/pytorch/models/sdxl-prompt-encoder-tank/test_cases.json
saienduri: Hmm yeah, this is probably easier to decode/update

.github/workflows/test_iree.yml
ScottTodd: Yeah I've been using 4.

.github/workflows/test_iree.yml
saienduri: I was just thinking in the future, if we want to run the flow on different machines, but not needed now

iree_tests/configs/config_sdxl_cpu_llvm_task.json
saienduri: Looks like the second one is, but removed the first one

saienduri:
@ScottTodd is this good to merge now?

ScottTodd (approved):

iree_tests/pytorch/models/sdxl-prompt-encoder-tank/test_cases.json
ScottTodd: Docs would still be nice.

PR #156, opened 2024-04-04
Adds a model.py for opt-125m

zjgarvey:
1. opt-125m is not in azure storage yet. I'll see if I can figure out how to add it. 
2. using past_sequence_length != 0 results in an onnx runtime error that looks virtually identical to this issue: [onnx#6130](https://github.com/microsoft/onnxruntime/issues/6130). I'm not sure how the onnx file was generated, but perhaps there is an option mentioned in that thread which can fix the issue. 
3. using past_sequence_length = 0 will successfully run in onnx, but fails in torch-mlir due to shape issues present in the torch-onnx-mlir ir (possible importer issue).


e2eshark/onnx/models/opt-125m/model.py
kumardeepakamd, line 1: Could you please rename opt-125M to opt-125M-awq so that if opt-12M-gptq or opt-125M-vaiq comes later, we can identify it better. The rest looks good.

kumardeepakamd (approved):

e2eshark/onnx/models/opt-125m/model.py
zjgarvey: will do

PR #160, opened 2024-04-11
Adds an e2e test for DQL -> MatMulInteger seen in Opt

zjgarvey:
This is to test some work I'm doing on quantized batch matmul in torch-mlir to get Opt-125m to lower to linalg. 

For now, I'm focusing on the one case I'm seeing in opt-125m which is rank3 x rank2 -> rank3 MatMulInteger ops.


kumardeepakamd (approved):
Great to see combination tests being added. Key to building quality efficiently.

PR #163, opened 2024-04-12
Upgrade Onnx Model Loading

PhaneeshB:
- adds support to load models from private storage if user is logged in to Azure on CLI with `az login` (and has permissions)
- failed to download / Unzip will not lead to stopping the tests entirely, only tests for specific models will fail with model.onnx file not found 
- filters testsList to contain only `models` tests (no operator / combo)


e2eshark/tools/aztestsetup.py
```diff
@@ -6,8 +6,12 @@
 
 import os, sys, argparse, shutil, zipfile
 from azure.storage.blob import ContainerClient
+from azure.core.exceptions import ResourceNotFoundError
 from pathlib import Path
+from zipfile import ZipFile
 
+PRIVATE_CONN_STRING = os.environ.get("PRIVATE_CONN_STRING", default="")
```
kumardeepakamd: Environment variables can be hard to keep track of. We can have some as needed but wherever possible may be we should just have a switch with a default?

kumardeepakamd: Also, if having env variable makes sense for a particular use case, may I advise to have a naming convention that makes it easier to remember? E2ESHARK_AZ_PRIVSTORAGE_BLOBNAME -- something like that shorter but very clear what it is...

kumardeepakamd: Looks like it is a connection setup. So, may be just call it: AZ_PRIVATE_CONNECTION and keep as env. We cannot make value of it public and can pass around internally for usage.

e2eshark/tools/aztestsetup.py
```diff
@@ -76,10 +100,47 @@ def download_onxx_model_from_azure_storage(cache_dir, testList):
...
+
+        if try_private:
+            print("Trying download from private storage")
+            try:
+                download_azure_blob(priv_account_url, priv_container_name, blob_name, dest_file)
+            except Exception as e:
+                print(f"Unable to download model for {model}.\nError - {type(e).__name__}")
+
+
+def pre_test_onnx_models_setup(testsList, cache_path, script_dir):
```
kumardeepakamd, line 121: Nit picking, but important to name identifiers to say what it does for productivity and keeping hygiene. Looks like this function downloads model too. Can have a more appropriate name?

kumardeepakamd (approved):
Good to be merged, but please address the comments.

e2eshark/tools/aztestsetup.py
saienduri: Done

e2eshark/tools/aztestsetup.py
saienduri: Sure, changed to pre_test_onnx_models_azure_download

PR #164, opened 2024-04-12
Refactor how config and test_cases.json files are handled in iree_tests.

ScottTodd:
Assorted quality of life improvements to iree_tests before we scale up to having more tests with real weights downloaded from Azure. These changes should also help as we move towards having files cached on CI test runners.

* Collapse/simplify how remote files are listed from
  ```json
        "remote_file_groups": [
          {
            "azure_account_url": "https://sharkpublic.blob.core.windows.net",
            "azure_container_name": "sharkpublic",
            "azure_base_blob_name": "sai/sd-unet-tank/",
            "files": [
              "inference_input.0.bin",
              "inference_input.1.bin",
  ```
  to
  ```json
        "remote_files": [
          "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd-unet-tank/inference_input.0.bin",
          "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd-unet-tank/inference_input.1.bin",
  ```
  (That was premature optimization, [KISS](https://en.wikipedia.org/wiki/KISS_principle), etc. etc.)
* Skip downloading remote files if they exist on disk with a matching MD5 hash (requires the files to have MD5 hashes computed in Azure -- some files are missing them)
* Relax JSON parsing code to tolerate omitted fields (e.g. instead of requiring `"skip_run_tests": [],`, you can now omit it)
* Add a `"file_format": "test_cases_v0",` marker to JSON files and look for that instead of a specific file name
* Print file sizes while downloading, e.g.
  ```bash
  $ python download_remote_files.py

  Processing future-pytorch-models\llama-tank\test_cases.json
    Downloading 'llama-tank.mlirbc' (128.44 MiB) to 'future-pytorch-models\llama-tank'
  Processing future-pytorch-models\sd-clip-tank\test_cases.json
    Skipping 'inference_input.0.bin' download (local MD5 hash matches)
    Skipping 'inference_output.0.bin' download (local MD5 hash matches)
    Skipping 'inference_output.1.bin' download (local MD5 hash matches)
    Downloading 'real_weights.irpa' (469.46 MiB) to 'future-pytorch-models\sd-clip-tank'
  ```


iree_tests/download_remote_files.py
```diff
@@ -54,11 +121,17 @@ def download_for_test_case(test_dir: Path, test_case_json: dict):
     )
     args = parser.parse_args()
 
-    for test_cases_path in (THIS_DIR / args.root_dir).rglob("test_cases.json"):
-        print(f"Processing {test_cases_path.relative_to(THIS_DIR)}")
+    # TODO(scotttodd): build list of files _then_ download
+    # TODO(scotttodd): report size needed for requested files and size available on disk
```
ScottTodd, line 125: Current file sizes, FYI: 
image

iree_tests/download_remote_files.py
```diff
@@ -4,45 +4,112 @@
...
+    with BlobClient(
+        account_url,
+        container_name,
+        blob_name,
+        max_chunk_get_size=1024 * 1024 * 32,  # 32 MiB
+        max_single_get_size=1024 * 1024 * 32,  # 32 MiB
+    ) as blob_client:
+        blob_properties = blob_client.get_blob_properties()
+
+        local_file_path = test_dir / remote_file_name
```
saienduri, line 77: Can we also make the downloading change to prepend with the environment variable here (local_file_path = IREE_TEST_FILES / test_dir / remote_file_name). IREE_TEST_FILES = os.environ.get("IREE_TEST_FILES", default=""). I think that's all needed. I'll update all the tests in another PR, but just want all the downloading changes to be made here.

saienduri:
Everything looks great to me! Just that one addition, and we should be good

iree_tests/download_remote_files.py
ScottTodd: Could do that in this PR or a follow-up. I'd prefer to keep this change smaller since it already touches a few things at once.

For that method of working with caching, we could also just populate the cache by

1. Checkout test suite repo
2. Run download script
3. Copy repo folder to cache directory

No need to teach the download script itself about other paths then...

(but also, whatever works - we don't need to overengineer this)

saienduri: Ok cool, let's just merge this for now then. I can figure out caching from there

saienduri (approved):

iree_tests/download_remote_files.py
```diff
@@ -4,45 +4,112 @@
...
+        if check_azure_remote_file_matching_local(blob_properties, local_file_path):
+            print(f"  Skipping '{remote_file_name}' download (local MD5 hash matches)")
+            return
+
+        size_str = human_readable_size(blob_properties.size)
+        print(f"  Downloading '{remote_file_name}' ({size_str}) to '{relative_dir}'")
```
ScottTodd, line 83: Going to iterate on the logging here. This output is buffered when we want it to be streamed / flushed more regularly


https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8708216214/job/23885060522?pr=164#step:6:1
```
Tue, 16 Apr 2024 15:19:19 GMT  python3 iree_tests/download_remote_files.py --root-dir pytorch/models
...
Tue, 16 Apr 2024 15:38:09 GMT Processing pytorch/models/opt-125M/test_cases.json
```

saienduri:
I was messing around with the caching yesterday (building on your support). For some reason (https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8703496936/job/23869795356), the download doesn't get skipped for real_weight.irpa for prompt encoder, scheduled unet, and vae decode. Can you check what's going on with the md5 hashing with any of these files? (maybe the size of the file is throwing it off)

ScottTodd:
> I was messing around with the caching yesterday (building on your support). For some reason (https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8703496936/job/23869795356), the download doesn't get skipped for real_weight.irpa for prompt encoder, scheduled unet, and vae decode. Can you check what's going on with the md5 hashing with any of these files? (maybe the size of the file is throwing it off)

Yeah I was just writing up a comment on that. I added some more complete logging: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8709108576/job/23888057849?pr=164#step:6:47 . Looks like depending on how the file is uploaded to Azure a md5 hash may or may not be calculated for us: https://learn.microsoft.com/en-us/answers/questions/282572/md5-hash-calculation-for-large-files

saienduri:
Ahh okay that makes sense. Looks like the md5 hashes are being generated, but in a different way that is not defined for the larger files. Two things we can do here: We can either add a text file with every file we download with the md5 hash in it for the blob (sdxl_vae_decode/real_weights_md5.txt). This way we just match the md5s that are generated by Azure however they are done without doing any hashing on our end. Or, we just have the model downloading add a --update-cache switch. If this switch is not set, we just check if the file exists. If it exists, we skip the downloading and don't worry about if we have the latest. And, we can add a CI that updates the cache nightly or weekly (whatever we think is more neccessary).

ScottTodd:
Some of the large files I uploaded have MD5 hashes set. We could just change how we upload files or upload then set the MD5 hashes explicitly. I think I can backfill the existing files.

I was thinking we could have regular jobs start by downloading into the cache (to pick up any newly added files without waiting for an explicit cache populate job). If we assume that files don't change frequently then we only need to store 1 local version per file -- otherwise we'd want some smarter caching logic with an eviction policy.

>  We can just have the model downloading file add a --update-cache switch (can add this with the caching PR). If this switch is not set, we just check if the file exists. If it exists, we skip the downloading and don't worry about if we have the latest.

Definitely a few things we could try with flags and failure modes, yeah. We should optimize for downstream developer experience, meaning:
* Predictable CI times - not accidently spending 1 hour downloading files
* Predictable/consistent sets of tests - when we pin to a SHARK-TestSuite commit, we should expect all the tests at that commit to run

That might mean adding a `--strict` style flag that checks for the cache being populated and fails if it isn't. I'm skeptical about accepting a stale cache though.

PR #165, opened 2024-04-15
Updated e2eshark CI

saienduri:
This commit adds a nightly run that will run and generate reports for all the onnx and pytorch models. Then, it will upload them here: https://github.com/nod-ai/e2eshark-reports.


.github/workflows/test_e2eshark.yml
```diff
@@ -0,0 +1,100 @@
...
+          pip install --upgrade pip
+          pip install -r ./e2eshark/requirements.txt
+        working-directory: ./test-suite
+
+      - name: Run Onnx Mode
+        run: |
+          source ${E2E_VENV_DIR}/bin/activate
+          cd e2eshark
+          free -mh
+          HF_TOKEN=${{ secrets.HF_TOKEN }} AZ_PRIVATE_CONNECTION=${{ secrets.AZ_PRIVATE_CONNECTION }} python3.11 ./run.py -r ./test-onnx --report --cachedir ~/huggingface_cache --mode onnx -f onnx pytorch -g models --tolerance .001 .001 --cleanup --postprocess --ci -v
```
ScottTodd, line 50: Split this onto multiple lines using `\`?

.github/workflows/test_e2eshark.yml
```diff
@@ -0,0 +1,100 @@
...
+        working-directory: ./test-suite
+
+      - name: Steup turbine python venv
```
ScottTodd, line 53: typo
```suggestion
      - name: Setup turbine python venv
```

.github/workflows/test_e2eshark.yml
```diff
@@ -0,0 +1,100 @@
...
+          path: test-suite
+
+      - name: "Setup e2eshark python venv"
+        run: |
+          python3.11 -m venv ${E2E_VENV_DIR}
+          pip install --upgrade pip
+          pip install -r ./e2eshark/requirements.txt
+        working-directory: ./test-suite
+
+      - name: Run Onnx Mode
```
ScottTodd, line 45: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8895268824/job/24425273404#step:4:577
browsing through the logs, gpt2-xl looks pretty slow (sorta expected given the name):
image

e2eshark/run.py
ScottTodd, line 1: Logs have some `reoprt` typos in them:

https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8895268824/job/24425273404#step:4:787
```
14:08:51 GMT
Generated status reoprt /mnt/share/amdsharks/actions-runner/_work/SHARK-TestSuite/SHARK-TestSuite/test-suite/e2eshark/test-onnx/statusreport.md
Tue, 30 Apr 2024 14:08:51 GMT
Generated time reoprt /mnt/share/amdsharks/actions-runner/_work/SHARK-TestSuite/SHARK-TestSuite/test-suite/e2eshark/test-onnx/timereport.md
Tue, 30 Apr 2024 14:08:51 GMT
Generated summary reoprt /mnt/share/amdsharks/actions-runner/_work/SHARK-TestSuite/SHARK-TestSuite/test-suite/e2eshark/test-onnx/summaryreport.md
```

e2eshark/run.py
```diff
@@ -843,7 +843,7 @@ def runTestUsingClassicalFlow(args_tuple):
     if args.verbose:
         print(f"Running classical flow for test {testName}")
     ireeInputType = "onnx"
-    if SHARED_TORCH_MLIR_BUILD:
+    if SHARED_TORCH_MLIR_BUILD or mode=="turbine":
         ireeInputType = "torch"
```
ScottTodd, line 847: You can also omit `--iree-input-type` and let the auto pipeline detect the input type based on the input program. I recommend that, especially since there is code like this that tries to be explicit but then gets the type wrong...

e2eshark/turbine-requirements.txt
```diff
@@ -0,0 +1,8 @@
+tabulate
+simplejson
+ml_dtypes
+sentencepiece
+accelerate
+auto-gptq
+optimum
+azure-storage-blob
```
ScottTodd, line 8: Some of these are already in https://github.com/nod-ai/SHARK-TestSuite/blob/main/e2eshark/requirements.txt. Want to split this into `ci-requirements.txt` and include just the extra packages? (Wouldn't want that file to pin to a specific version of one of these packages and that to possibly get out of sync with this file)

.github/workflows/test_e2eshark.yml
```diff
@@ -0,0 +1,100 @@
...
+    needs: [e2eshark]
+    runs-on: nodai-amdgpu-mi250-x86-64
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+        with:
+          repository: nod-ai/e2eshark-reports
+          ref: 'main'
+          token: ${{ secrets.E2ESHARK_GITHUB_TOKEN }}
+          path: e2eshark-reports
```
ScottTodd, line 87: It looks like this is depending on running on the same machine with some files persisting. If there are multiple runners matching this label then that will not always be true. If there are multiple actions running then the files may be overwritten between when the previous job runs and when this job runs. Instead, you could either
* put all the jobs together
* pass report files between jobs: https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow

e2eshark/run.py
saienduri: Oh nice! Yeah that's a lot better

.github/workflows/test_e2eshark.yml
saienduri: Cool, this is good to know (will be helpful with more intense jobs that need to pass artifacts). Here, I will just combine them because the upload is really quick and can just be combined.

ScottTodd: You can also pass files between jobs using cloud storage (especially if they are too large for GitHub Artifacts).
We do that in IREE today: 
* https://github.com/iree-org/iree/blob/e4b5a9350aee7cfafefb440a7f01b3d376842d92/.github/workflows/benchmark_execution.yml#L192-L193
* https://github.com/iree-org/iree/blob/e4b5a9350aee7cfafefb440a7f01b3d376842d92/.github/workflows/ci.yml#L986-L1001

(but I've wanted to move some of those to workflow artifacts using standard github infra if possible)

ScottTodd: It will also help when running across multiple machines (e.g. w7900, MI250, MI300). Distribute tests across workers -> wait on all workers and aggregate the results

e2eshark/turbine-requirements.txt
saienduri: Actually the point of this was to remove some of the requirements in the e2eshark requirements because it conflicts with the turbine requirements and causes pip problems. I will see if there is a way to use a requirements.txt pip install but ignore some of the requirements

ScottTodd: Oh... interesting. If you have logs or repro instructions someone else (me / others) could help debug that too.

kumardeepakamd (approved):
Followed the conversation.  Lgtm. Go ahead and get it in and iterate if needed.

e2eshark/turbine-requirements.txt
saienduri: This won't be a problem in the future (nothing we should spend time debugging). It just has to do with different torch versions currently between e2eshark and turbine and the effects of that. Once torch 2.3 is in a good state for turbine, I will switch it over and we should be good :)

.github/workflows/test_e2eshark.yml
saienduri: Yeah, we can leave it for now given it's a nightly

.github/workflows/test_e2eshark.yml
saienduri: Made changes for data passing between jobs

PR #166, opened 2024-04-15
Add onnx models 2 and ghsot zip fix

PhaneeshB:
- fixes creation of ghost zip file when download fails 
- adds support for remaining onnx models to testsuite


saienduri (approved):
LGTM

Issue #169, opened 2024-04-16
Speed up iree_tests for large remote files by adding a caching layer on CI machines


ScottTodd:
This repo on its own has this directory structure:

```
SHARK-TestSuite/
  iree_tests/
    conftest.py
    onnx/
    pytorch/
      models/
        opt-125M/
          inference_input.0.bin
          model.mlirbc
          real_weights_data_flags.txt
        resnet50/
          model.mlirbc
```

The .bin and .mlirbc files are large enough that many are stored remotely (e.g. on Azure). For CI machines, we want those files to be cached locally so jobs don't need to redownload them every run (using bandwidth and taking 20+ minutes).

Caching can be tricky, so let's try to start simple and add features as needed. There may be an off the shelf caching layer or tool we can use if it gets to that point.

Brainstorming a few approaches:

1. Maintain a parallel directory structure in `~/iree_tests_cache/`:

    ```
    pytorch/
      models/
        opt-125M/
          inference_input.0.bin
    ```

    Run pytest from the test repo, redirecting test flagfile paths using an environment variable:

    ```diff
    - --input=1x8xi64=@inference_input.0.bin
    + --input=1x8xi64=@${IREE_TESTS_ROOT}/pytorch/models/opt-125M/inference_input.0.bin
    ```

    with that environment variable defaulting to the repository path itself and set to the cache dir on CI machines.

    That may not be portable across systems (Windows may want `%IREE_TESTS_ROOT%/`?).

2. Maintain a parallel directory with cached files, copy the test repo over it, then run pytest _from the cache dir_.

    This would pollute the cache with other files but would require minimal configuration changes and file move operations.

3. Maintain a parallel directory with cached files, copy files from the cache directory into the test repo, then run pytest _from the repo_.

    This would keep the cache clean, but would involve copying more files on the local disk (machines would then have two copies of large files)

4. Same as 3, but with symlinks. Maybe add a `setup_cache_symlinks.py` script to help? Could also look at overlay file systems / symlink trees (https://unix.stackexchange.com/a/179404) 

---

* All of these approaches need a way to initially populate and then update the cache.
* Ideally the test runner itself wouldn't need to know about caching at all. I'd prefer symlinking or copying files over adding new environment variables for that reason.

saienduri:
[...]

PR #170, opened 2024-04-16
Add caching support for model testing

saienduri:
This commit adds caching support for model testing. This brings the download time from ~20 mins to 23 seconds. With these changes, it is feasible to add model testing as a presubmit in the iree repo which was the reason for these changes. 

Description of what's going on:  
First, we use an environment variable passed in (IREE_TEST_FILES) to determine if we even want to use a cache or not (don't want to make it a requirement). In CI, we set this because we want a persistent runner cache. If this is set, we download files to the cache location specified and symlink it to the testing directory. Otherwise, we just directly download into the repo, and it won't be using any caching mechanism. The test runner doesn't need to know about caching at all, and even initially, it will just create the cache and download there if it doesn't exist. There is no manual interaction that needs to take place for any system or person that wants to use this.

Example of actions run first time and second time (runner knows nothing about caching, no manual setup):
https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8711532825
https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8711798501

Final model testing job time: ~5 mins

The machine in the iree repo is even faster, so can expect lower times :)

Fixes https://github.com/nod-ai/SHARK-TestSuite/issues/169

### DIFF
--- a/iree_tests/download_remote_files.py
+++ b/iree_tests/download_remote_files.py
@@ -12,8 +12,10 @@ import logging
 import mmap
 import pyjson5
 import re
+import os
 
 THIS_DIR = Path(__file__).parent
+REPO_ROOT = Path(__file__).parent.parent
 logger = logging.getLogger(__name__)
 
 
@@ -81,7 +83,18 @@ def download_azure_remote_file(test_dir: Path, remote_file: str):
         blob_size_str = human_readable_size(blob_properties.size)
         remote_md5 = get_remote_md5(remote_file, blob_properties)
 
-        local_file_path = test_dir / remote_file_name
+        cache_location = os.getenv("IREE_TEST_FILES", default="")
+        if cache_location == "":
+            os.environ["IREE_TEST_FILES"] = str(REPO_ROOT)
+            cache_location = REPO_ROOT
+        if cache_location == REPO_ROOT:
+            local_dir_path = test_dir
+            local_file_path = test_dir / remote_file_name
+        else:
+            cache_location = os.path.expanduser(cache_location)
+            local_dir_path = Path(cache_location) / "iree_tests" / relative_dir
+            local_file_path = Path(cache_location) / "iree_tests" / relative_dir / remote_file_name
+
         local_md5 = get_local_md5(local_file_path)
 
         if remote_md5 and remote_md5 == local_md5:
@@ -89,6 +102,7 @@ def download_azure_remote_file(test_dir: Path, remote_file: str):
                 f"  Skipping '{remote_file_name}' download ({blob_size_str}) "
                 "- local MD5 hash matches"
             )
+            os.symlink(local_file_path, test_dir / remote_file_name)
             return
 
         if not local_md5:
@@ -102,9 +116,13 @@ def download_azure_remote_file(test_dir: Path, remote_file: str):
                 f"to '{relative_dir}' (local MD5 does not match)"
             )
 
+        if not os.path.isdir(local_dir_path):
+            os.makedirs(local_dir_path)
         with open(local_file_path, mode="wb") as local_blob:
             download_stream = blob_client.download_blob(max_concurrency=4)
             local_blob.write(download_stream.readall())
+        if str(cache_location) != str(REPO_ROOT):
+            os.symlink(local_file_path, test_dir / remote_file_name)
 
 
 def download_generic_remote_file(test_dir: Path, remote_file: str):
--- a/iree_tests/download_remote_files.py
+++ b/iree_tests/download_remote_files.py
@@ -91,9 +91,9 @@ def download_azure_remote_file(test_dir: Path, remote_file: str):
             local_dir_path = test_dir
             local_file_path = test_dir / remote_file_name
         else:
-            cache_location = os.path.expanduser(cache_location)
-            local_dir_path = Path(cache_location) / "iree_tests" / relative_dir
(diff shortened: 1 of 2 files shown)


iree_tests/download_remote_files.py
```diff
@@ -81,14 +83,26 @@ def download_azure_remote_file(test_dir: Path, remote_file: str):
         blob_size_str = human_readable_size(blob_properties.size)
         remote_md5 = get_remote_md5(remote_file, blob_properties)
 
-        local_file_path = test_dir / remote_file_name
+        cache_location = os.getenv("IREE_TEST_FILES", default="")
+        if cache_location == "":
+            os.environ["IREE_TEST_FILES"] = str(REPO_ROOT)
+            cache_location = REPO_ROOT
+        if cache_location == REPO_ROOT:
```
ScottTodd, line 90: String and Path handling here could be tricky. We might want to convert the string coming from the environment variable to a canonical path before handling it further. 

Could use https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve for that

iree_tests/download_remote_files.py
```diff
@@ -81,14 +83,26 @@ def download_azure_remote_file(test_dir: Path, remote_file: str):
...
+            local_dir_path = test_dir
+            local_file_path = test_dir / remote_file_name
+        else:
+            cache_location = os.path.expanduser(cache_location)
```
ScottTodd, line 94: Related to my other comment, I think `resolve()` covers more cases than just `expanduser`

iree_tests/download_remote_files.py
```diff
@@ -81,14 +83,26 @@ def download_azure_remote_file(test_dir: Path, remote_file: str):
...
+            local_file_path = Path(cache_location) / "iree_tests" / relative_dir / remote_file_name
+
         local_md5 = get_local_md5(local_file_path)
 
         if remote_md5 and remote_md5 == local_md5:
             logger.info(
                 f"  Skipping '{remote_file_name}' download ({blob_size_str}) "
                 "- local MD5 hash matches"
             )
+            os.symlink(local_file_path, test_dir / remote_file_name)
```
ScottTodd, line 105: Let's add some logging for the symlink setup too. (We want enough information in the logs for future debugging)

iree_tests/download_remote_files.py
```diff
@@ -102,9 +116,13 @@ def download_azure_remote_file(test_dir: Path, remote_file: str):
                 f"to '{relative_dir}' (local MD5 does not match)"
             )
 
+        if not os.path.isdir(local_dir_path):
+            os.makedirs(local_dir_path)
         with open(local_file_path, mode="wb") as local_blob:
             download_stream = blob_client.download_blob(max_concurrency=4)
             local_blob.write(download_stream.readall())
+        if str(cache_location) != str(REPO_ROOT):
+            os.symlink(local_file_path, test_dir / remote_file_name)
```
ScottTodd, line 125: Could add logging here

iree_tests/download_remote_files.py
ScottTodd, line 1: I'm wondering if we should split the scripts into some separate scripts / stages so they can be used piecemeal, like `download_remote_files.py`, `populate_cache.py`, `configure_cache.py`. That can be done in a follow-up.

Steady state CI jobs (with cache):
* `download_remote_files.py` or `populate_cache.py` (downloads into the cache)
* `configure_cache.py` (sets up symlinks)

CI maintenance (e.g. first time adding a runner):
* `download_remote_files.py` or `populate_cache.py` (downloads into the cache)

Local runs:
* `download_remote_files.py`

Basically trying to avoid overloading a single script with environment variables and flags -- it should be obvious which script(s) to run in each scenario.

ScottTodd (approved):
Cool! This looks great to get us going.

iree_tests/download_remote_files.py
```diff
@@ -103,6 +103,9 @@ def download_azure_remote_file(test_dir: Path, remote_file: str):
                 "- local MD5 hash matches"
             )
             os.symlink(local_file_path, test_dir / remote_file_name)
+            logger.info(
+                f"  Created symlink for '{local_file_path}' to '{test_dir / remote_file_name}' "
+            )
```
ScottTodd, line 108: nit: can drop trailing ` ` (space). I included those in other sections since the string continued on the next line

iree_tests/download_remote_files.py
```diff
@@ -123,6 +126,9 @@ def download_azure_remote_file(test_dir: Path, remote_file: str):
             local_blob.write(download_stream.readall())
         if str(cache_location) != str(REPO_ROOT):
             os.symlink(local_file_path, test_dir / remote_file_name)
+            logger.info(
+                f"  Created symlink for '{local_file_path}' to '{test_dir / remote_file_name}' "
+            )
```
ScottTodd, line 131: same here

ScottTodd (approved):

iree_tests/download_remote_files.py
saienduri: Yeah, I can see the value in this follow-up instead of having one beast of a file. Makes it easier to understand what is going on and reuse utilities implemented in other places.

iree_tests/download_remote_files.py
saienduri: Ah I see. I thought it was some sort of styling format for logging, thanks

PR #172, opened 2024-04-17
Adds option to run opt-125M-awq as a static model

zjgarvey:
I'm not sure how I feel about saving the dynamic model to the test bin. Any suggestions?

This also fixes an issue with a random int64 input always being zero.

### DIFF
--- a/e2eshark/onnx/models/opt-125M-awq/model.py
+++ b/e2eshark/onnx/models/opt-125M-awq/model.py
@@ -1,6 +1,7 @@
 import numpy, torch, sys
 import onnxruntime
 import onnx
+from onnxruntime.tools.onnx_model_utils import make_dim_param_fixed, fix_output_shapes
 
 # import from e2eshark/tools to allow running in current dir, for run through
 # run.pl, commutils is symbolically linked to allow any rundir to work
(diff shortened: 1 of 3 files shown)


e2eshark/onnx/models/opt-125M-awq/model.py
```diff
@@ -16,34 +16,92 @@
...
+pv_zip = zip(dim_params, dim_values)
+pv = dict(pv_zip)
+
+model = onnx.load("model.onnx")
+isDynamic = bool(model.graph.input[0].type.tensor_type.shape.dim[0].dim_param)
+
+if run_as_static and isDynamic:
+    print("Saving dynamic model...")
+    onnx.save(model, "dynamic_model.onnx")
+    print("\t dynamic model saved as dynamic_model.onnx")
```
ScottTodd, line 37: May want to use `logging` (https://docs.python.org/3/library/logging.html) with a scoped logger if adding so many `print` calls. That way developers could redirect logging, change which modules log at what verbosity, and see which logs come from which file.

e2eshark/onnx/models/opt-125M-awq/model.py
```diff
@@ -16,34 +16,92 @@
...
+
+    print("Setting model dim_params:")
+    for p, v in pv_zip:
+        make_dim_param_fixed(model.graph, p, v)
+        print("\t", p, " = ", v)
```
ScottTodd, line 42: On the topic of logging, could use f-strings (https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals). As written, this will have two spaces around the `=`.
```suggestion
        print(f"\t {p} = {v}")
```

e2eshark/onnx/models/opt-125M-awq/model.py
```diff
@@ -16,34 +16,92 @@
...
+    "sequence_length",
+    "past_sequence_length",
+    "past_sequence_length + 1",
+)
+dim_values = (12, 7, 0, 1)
+pv_zip = zip(dim_params, dim_values)
+pv = dict(pv_zip)
+
+model = onnx.load("model.onnx")
+isDynamic = bool(model.graph.input[0].type.tensor_type.shape.dim[0].dim_param)
```
ScottTodd, line 32: match naming style https://peps.python.org/pep-0008/#function-and-variable-names
```suggestion
is_dynamic = bool(model.graph.input[0].type.tensor_type.shape.dim[0].dim_param)
```

e2eshark/onnx/models/opt-125M-awq/model.py
```diff
@@ -16,34 +16,92 @@
 # to form runmodel.py in the rundirectory which is then taken
 # through flow
 
-# set some variable parameters (dynamic dims)
-batch_size = 3 
-sequence_length = 7 
-past_sequence_length = 0 # setting this parameter to anything other than 0 results in an error
+# If you want to run the test with a static model.onnx, set run_as_static = True
+run_as_static = False
```
ScottTodd, line 20: Please clarify what specifically is static or dynamic about the model.

Also, is this worth making configurable via `argparse` (https://docs.python.org/3/library/argparse.html)?

e2eshark/onnx/models/opt-125M-awq/model.py
ScottTodd, line 37: > I'm not sure how I feel about saving the dynamic model to the test bin. Any suggestions?

One thing we've been setting up in [`iree_tests/`](https://github.com/nod-ai/SHARK-TestSuite/tree/main/iree_tests) is an external cache directory that you can set via an environment variable (or CLI flag). For individual developers, a directory structure under the repository itself is usually fine (as long as it is clearly marked, added to `.gitignore`, and is easy to clean). For CI jobs and on shared machines, prefer to store and reference large files that don't change test run to test run in an external cache directory.

e2eshark/onnx/models/opt-125M-awq/model.py
```diff
@@ -16,34 +16,92 @@
...
+
+    print("Setting model dim_params:")
+    for p, v in pv_zip:
+        make_dim_param_fixed(model.graph, p, v)
+        print("\t", p, " = ", v)
+
+    fix_output_shapes(model)
+    print("Overwriting file contents of model.onnx...")
+    onnx.save(model, "model.onnx")
+    print("\t model.onnx is now a static model.")
```
ScottTodd, line 47: (was curious and did some research)
if you want to get fancy with indented logging (`\t ...`), https://python-log-indenter.readthedocs.io/en/latest/usage.html is an option (requires an extra package though)

zjgarvey:
@ScottTodd Thanks for all of the helpful tips. I'll check out the suggestions and make some changes.

e2eshark/onnx/models/opt-125M-awq/model.py
zjgarvey: Hm, I'll need to look at how that is set up. This approach, in general, might be more annoying from a CI perspective than just adding a separate test entirely for a static model.

zjgarvey:
@ScottTodd I tried to find a good way to pass an arg from run.py to model.py, and decided to roll the model.py contents into a function, and try calling that function in [e2eshark/tools/stubs/onnxmodel.py](https://github.com/nod-ai/SHARK-TestSuite/pull/172/files#diff-9ff456f427974d0f430805a9ce18f142150e150ac4c795354188ad3694d0765f). This way it won't run the model until the argparsing happens. For other models, it should just throw an exception and either warn the user if they mistakenly passed --run_as_static, or just continue with the original E2ESHARK_CHECK business. 

If you have a chance to check it out again, I'd appreciate it.

ScottTodd:
> @ScottTodd I tried to find a good way to pass an arg from run.py to model.py, and decided to roll the model.py contents into a function, and try calling that function in [e2eshark/tools/stubs/onnxmodel.py](https://github.com/nod-ai/SHARK-TestSuite/pull/172/files#diff-9ff456f427974d0f430805a9ce18f142150e150ac4c795354188ad3694d0765f). This way it won't run the model until the argparsing happens. For other models, it should just throw an exception and either warn the user if they mistakenly passed --run_as_static, or just continue with the original E2ESHARK_CHECK business.
> 
> If you have a chance to check it out again, I'd appreciate it.

Not sure I'll have immediately actionable feedback there... the run.py, magic global variables, and stubs files setup is very strange to me, and it's one of the things that motivated me to start a new directory using pytest and to draw a very clear line between e2eshark and iree_tests.

ScottTodd:
> > @ScottTodd I tried to find a good way to pass an arg from run.py to model.py, and decided to roll the model.py contents into a function, and try calling that function in [e2eshark/tools/stubs/onnxmodel.py](https://github.com/nod-ai/SHARK-TestSuite/pull/172/files#diff-9ff456f427974d0f430805a9ce18f142150e150ac4c795354188ad3694d0765f). This way it won't run the model until the argparsing happens. For other models, it should just throw an exception and either warn the user if they mistakenly passed --run_as_static, or just continue with the original E2ESHARK_CHECK business.
> > If you have a chance to check it out again, I'd appreciate it.
> 
> Not sure I'll have immediately actionable feedback there... the run.py, magic global variables, and stubs files setup is very strange to me, and it's one of the things that motivated me to start a new directory using pytest and to draw a very clear line between e2eshark and iree_tests.

Had some time to look at this again. I like this direction this moves in. Defining functions or classes in these files will make them more composable than having loose functions and global variables. See how ONNX has test cases defined, like https://github.com/onnx/onnx/blob/main/onnx/backend/test/case/node/concat.py. There is a base class that each test file implements, with some methods and functions to call that configure what should be tested. A test runner (https://github.com/onnx/onnx/blob/main/onnx/backend/test/runner/__init__.py) then loads test cases and chooses how to run them.

ScottTodd (approved):

PR #173, opened 2024-04-17
add onnx if operator test and modify run.py to support bool types

renxida:


kumardeepakamd (approved):
Good to have bool support added. Test looks good.

PR #175, opened 2024-04-17
Adds a model.py for llama-7b-awq

zjgarvey:
A few things:

1. I had to manually download the model.onnx.zip. I'm not sure why the downloader in run.py is failing to get it. 
2. The model uses external weights. However, I was still able to load and inspect the model with 
```python
import onnx
model = onnx.load("model.onnx", load_external_data=False)
```
3. The external data file needs to be named EXACTLY the same as when it was generated. This is the printout from the error:
```
  File "/home/zjgar/code/SHARK-TestSuite/e2eshark/onnx/models/llama2-7b-awq/model.py", line 91, in <module>
    session = onnxruntime.InferenceSession("model.onnx", None)
  File "/home/zjgar/code/SHARK-TestSuite/ts.venv/lib/python3.10/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 419, in __init__
    self._create_inference_session(providers, provider_options, disabled_optimizers)
  File "/home/zjgar/code/SHARK-TestSuite/ts.venv/lib/python3.10/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 483, in _create_inference_session
    sess.initialize_session(providers, provider_options, disabled_optimizers)
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Deserialize tensor onnx::MatMul_7290_quantized failed.GetFileLength for ./model_quantized.onnx.data failed:Invalid fd was supplied: -1
```
Which suggests it was named "model_quantized.onnx.data". This file should be added to onnx secure storage (and likely zipped because it will probably be big), otherwise we won't be able to run this model. With some adjustments, we could still debug the onnx->torch->linalg pipeline in the meantime.


kumardeepakamd (commented):
Looks like need a way to store and download the weights i.e. model_quantized.onnx.data to/from Azure Storage. Will upload the weights first and suggest a way to handle it for setting up at download.

zjgarvey:
> Looks like need a way to store and download the weights i.e. model_quantized.onnx.data to/from Azure Storage. Will upload the weights first and suggest a way to handle it for setting up at download.

I'll also need to rework run_as_static to match the decisions made in #172, so this PR should wait a bit anyway.

kumardeepakamd:
Now model.onnx.zip has model_quantized.onnx.data

PR #176, opened 2024-04-18
Onnx.if op test

renxida:
Simple op test for if. Also includes a nifty little helper function that would reduce error & duplication by avoiding re-specifying dtypes and shapes

```python
inputs = session.get_inputs()
# gets Z in outputs[0]
outputs = session.get_outputs()

def generate_input_from_node(node: onnxruntime.capi.onnxruntime_pybind11_state.NodeArg):
    if node.type == "tensor(float)":
        return numpy.random.randn(*node.shape).astype(numpy.float32)
    if node.type == "tensor(int)":
        return numpy.random.randint(0, 10000, size=node.shape).astype(numpy.int32)
    if node.type == "tensor(bool)":
        return numpy.random.randint(0, 2, size=node.shape).astype(bool)
    
input_dict = {
    node.name: generate_input_from_node(node)
    for node in inputs
}

output_list = [
    node.name
    for node in outputs
]
```

### DIFF
--- a/e2eshark/onnx/operators/If/model.py
+++ b/e2eshark/onnx/operators/If/model.py
@@ -26,11 +26,11 @@ E2ESHARK_CHECK = dict(E2ESHARK_CHECK_DEF)
 
 # condition has to be a float tensor
 condition = make_tensor_value_info('condition', TensorProto.BOOL, [1])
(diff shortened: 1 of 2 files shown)


kumardeepakamd (approved):
Looks good

kumardeepakamd:
Looks like there is conflict to be resolved, otherwise is fine.

PR #178, opened 2024-04-18
Added a second gridsampler testcase to test both align_corners modes

afalkenberg1:
gridsampler has a parameter to select how corners and edges are aligned. 
The two options are now implemented and therefore two cases are provided with this PR.


kumardeepakamd (approved):
Good to have two tests

PR #194, opened 2024-04-22
SDXL ROCm Testing

saienduri:
This PR adds ROCm Testing support for all the sdxl models. Main things that had to be added was support to add backend + model specific compile and runtime flags.


iree_tests/configs/config_gpu_rocm_models.json
```diff
@@ -0,0 +1,17 @@
...
+      "--iree-opt-const-eval=false",
+      "--verify=false",
+      "--iree-codegen-transform-dialect-library=/mnt/share/amdsharks/sai/specs/attention_and_matmul_spec.mlir"
+    ],
+    "iree_run_module_flags": [
+      "--device=hip"
+    ],
+    "skip_compile_tests": [],
+    "skip_run_tests": [],
+    "expected_compile_failures": ["resnet50"],
```
ScottTodd, line 15: nit: could put this on a new line to make future lines possible to add without unrelated diffs

```suggestion
    "expected_compile_failures": [
      "resnet50",
    ],
```

compare:
```diff
    "expected_compile_failures": [
+     "mobilenet",
      "resnet50",
    ],
```

with:
```diff
-   "expected_compile_failures": ["resnet50"],
+   "expected_compile_failures": [
+     "mobilenet",
+     "resnet50",
+   ],
```

iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/sdxl_scheduled_unet_pipeline_fp16_rocm.vmfb
ScottTodd, line 1: What is different between this `fp16_rocm.vmfb` file and the `fp16_cpu.vmfb` file? We shouldn't need a separate binary file for each target/configuration. Anyone downstream (e.g. in https://github.com/nod-ai/iree-amd-aie) should be able to run these tests without modifying the upstream test definitions

iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/real_weights_data_flags_rocm_llvm_task.txt
ScottTodd, line 1: What's the "llvm_task" part of this file name?

iree_tests/conftest.py
```diff
@@ -280,13 +283,25 @@ def collect(self):
             #   try pytest-dependency or pytest-depends
             for test_case in test_cases:
                 test_name = config_name + "_" + test_case.name
+                # use backend specific runtime flagfile if it exists (don't just add because order matters)
+                config_specific_flagfile = test_directory / (test_case.runtime_flagfile.split(".")[0] + "_" + config_name + ".txt")
```
ScottTodd, line 287: This requires that downstream config files (e.g. https://github.com/iree-org/iree/blob/main/build_tools/pkgci/external_test_suite/pytorch_models_cpu_llvm_task.json) use names matching the upstream flagfiles (e.g. `iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/real_weights_data_flags_cpu_llvm_task.txt`). That will make it harder to add new configurations -- especially with out of tree targets like https://github.com/nod-ai/iree-amd-aie . Downstream should be able to just reference the files provided by the test suite, it should not need to make changes to the test suite itself.

A few ideas:

## One config per model (when needed)

Add a new .json config file for each set of flags.

https://github.com/iree-org/iree/blob/main/build_tools/pkgci/external_test_suite/pytorch_models_cpu_llvm_task.json could be forked into `pytorch_model_sdxl_scheduled_unet_cpu_llvm_task.json`, with extra flags added as needed and then tested with `pytest SHARK-TestSuite/iree_tests/pytorch/models/sdxl-scheduled-unet`

With issues like https://github.com/iree-org/iree/issues/17137, I'm expecting we'll have a few configs per model/platform that we'll want to test, like "default flags", "performance flags (config A)", "performance flags (config b)".

## Redefine the config format

The config files are JSON at the moment to keep them simple. We can replace them with something more powerful like Python to add whatever logic we want. For example, see how test cases are defined in https://github.com/iree-org/iree/blob/main/experimental/regression_suite/tests/pregenerated/test_llama2.py . The ONNX test suite has 1000+ test cases, so I wanted to lean on something more automated like the JSON configs for it. For model tests, being explicit about each test case might be easier.

ScottTodd (changes requested):

iree_tests/conftest.py
ScottTodd: For redefining the config format, I believe we can write Python code that substitutes for the "collect" phase in pytest, replacing this: https://github.com/nod-ai/SHARK-TestSuite/blob/2dba6a8e4d56ca146e25077cfa96f82001d2bc33/iree_tests/conftest.py#L248-L296

Could try using `IreeCompileAndRunTestSpec` directly, or extract some parts of that file into fixtures (like https://github.com/iree-org/iree/blob/main/experimental/regression_suite/ireers/fixtures.py, which we could also just use directly) or other helper functions.

I think we can keep these parts of testing infrastructure pretty separable:
* Model files, parameters, inputs, outputs
* Automated test case discovery (can replace with explicit/manual test definitions)
* Device/configuration-specific flags
* Test running

ScottTodd: For this PR I'd probably start with one config per model though. (work with what we have for N=1 and N=2, maybe up to N=10... generalize and refactor to reduce duplication and make maintenance easier when N=3+).

iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/sdxl_scheduled_unet_pipeline_fp16_rocm.vmfb
saienduri: So, these are the vmfb versions of this: https://github.com/nod-ai/SHARK-Turbine/blob/main/models/turbine_models/custom_models/sdxl_inference/sdxl_pipeline_bench_f16.mlir. Do you have another idea on how to deal with this? We need this mlir to test the "scheduled" part of the unet lol.

iree_tests/conftest.py
saienduri: Ok yeah this all makes sense. Will create config files where needed for now.

iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/sdxl_scheduled_unet_pipeline_fp16_rocm.vmfb
ScottTodd: Hmmm... for SDXL, how about we just do what works. One .vmfb per config in this repo for now isn't that bad.

I haven't wrapped my head fully around how that is working, but a few ideas come to mind:

* Bundle the "scheduled" part into the model (single .mlir file), could be spliced together with the help of a script or compiler pass
* Teach the test runner about pipelines (multiple .mlir files to compile and link/run together)

iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/real_weights_data_flags_rocm_llvm_task.txt
saienduri: removed was just going based on cpu name

iree_tests/configs/config_gpu_rocm_models.json
```diff
@@ -0,0 +1,17 @@
+{
+    "config_name": "rocm_llvm_task",
+    "iree_compile_flags" : [
+      "--iree-hal-target-backends=rocm",
+      "--iree-rocm-target-chip=gfx90a",
+      "--iree-opt-const-eval=false",
+      "--verify=false",
```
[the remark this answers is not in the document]
saienduri: ok cool, these were originally given to me as the most minimal flags possible. But, hey, it works without it with no real perf difference :)

iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/sdxl_scheduled_unet_pipeline_fp16_rocm.vmfb
saienduri: Ok got it, yeah maybe we can teach the test runner about pipelines. Not sure how the single mlir file would work, but that would be cool. I've only ever seen one module defined in a mlir file, so not sure how this would look and work at runtime.

saienduri:
Ok this patch is finally good to go I think

iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/real_weights_data_flags.txt
ScottTodd, line 1: Did you mean to remove all flags from this file? Is this test case still running?

iree_tests/pytorch/models/sdxl-prompt-encoder-tank/specs/attention_and_matmul_spec.mlir
ScottTodd, line 1: Are all these `attention_and_matmul_spec.mlir` files the same? We could put them in a common location instead of duplicating.

I saw you trying a few commits to not treat these as test cases. Could add some extra logic to https://github.com/nod-ai/SHARK-TestSuite/blob/2dba6a8e4d56ca146e25077cfa96f82001d2bc33/iree_tests/conftest.py#L111-L113 that ignores file names matching a pattern (or included in a specific directory)

iree_tests/configs/config_sdxl_scheduled_unet_gpu_rocm.json
```diff
@@ -0,0 +1,34 @@
...
+      "--input=2x1280xf16=@inference_input.2.bin",
+      "--input=1xf16=@inference_input.3.bin",
+      "--expected_output=1x4x128x128xf16=@inference_output.0.bin",
+      "--expected_f16_threshold=0.8f"
+    ],
+    "skip_compile_tests": [],
+    "skip_run_tests": [],
+    "expected_compile_failures": [
+        "resnet50"
+    ],
```
ScottTodd, line 32: Let's drop resnet from this config, then only run pytest with this config using `sdxl` patterns

.github/workflows/test_iree.yml
```diff
@@ -115,3 +115,78 @@ jobs:
             --retries 2 \
             --retry-delay 5 \
             --durations=0 \
+            --config-files=iree_tests/configs/config_pytorch_models_cpu_llvm_task.json
+          pytest iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank \
```
ScottTodd, line 119: nit: I'd split this into its own job step

.github/workflows/test_iree.yml
```diff
@@ -115,3 +115,78 @@ jobs:
...
+            -rpfE \
+            -k real_weights \
+            --no-skip-tests-missing-files \
+            --capture=no \
+            --timeout=1200 \
+            --retries 2 \
+            --retry-delay 5 \
+            --durations=0 \
+            --config-files=iree_tests/configs/config_gpu_rocm_models.json
+          pytest iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank \
```
ScottTodd, line 178: same here - separate job step

iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank/real_weights_data_flags.txt
saienduri: Because we have them in the config files for each configuration in scheduled unet, we don't need any flags here

iree_tests/pytorch/models/sdxl-prompt-encoder-tank/specs/attention_and_matmul_spec.mlir
saienduri: The problem here is that it has to be in the same directory because relative paths are not working in the config or real weight files. I can add that logic to ignore these files so we don't need the specs folder though?

ScottTodd: > because relative paths are not working in the config or real weight files

I'd expect relative paths to work... hmmm..

> I can add that logic to ignore these files so we don't need the specs folder though?

Yeah. Maybe just check for a path ending with `_spec.mlir` to skip

.github/workflows/test_iree.yml
ScottTodd: I like being able to click through each part of a job and see the time breakdown for each:
image

This a case where we'd be optimizing for the reader of the CI logs and not the reader/writer of the workflow file

saienduri: Ah I see, ok cool

PR #200, opened 2024-04-22
Refactor how download_remote_files handles cache directories.

ScottTodd:
A few improvements for local workflows:

* Allow passing the cache directory via either an environment variable or a CLI flag
* Delete preexisting files before writing out a symlink to fix `FileExistsError: [WinError 183] Cannot create a file when that file already exists` on Windows
* Pull code into a helper function so it can be shared between Azure file downloading and generic (GCS/Huggingface/etc.) file downloading
* Update a few variable names and docstrings


saienduri (approved):
LGTM, thanks for cleaning up!

PR #201, opened 2024-04-23
Update config_onnx_cpu_llvm_sync

jinchen62:
Ran the tests with TOM torch-mlir at commit a8ba865fcab6475ff58c2beb14a1823fc25314c2


iree_tests/configs/config_onnx_cpu_llvm_sync.json
ScottTodd, line 1: TLDR: You can come back to this PR after updating the torch-mlir submodule in IREE and waiting for a nightly IREE release.

Sequencing is a bit tricky here. The CI job running in this repo uses the nightly IREE releases: https://github.com/nod-ai/SHARK-TestSuite/blob/2dba6a8e4d56ca146e25077cfa96f82001d2bc33/.github/workflows/test_iree.yml#L43-L53

You can test changes in this repo, but what ultimately matters is what gets distributed on PyPI, and that's based on the code used in IREE (and iree-turbine). When you update the torch-mlir submodule in IREE, the config files _in IREE_ will need to be updated: https://github.com/iree-org/iree/tree/main/build_tools/pkgci/external_test_suite. Those are intentionally decoupled from the config file here, so downstream projects (like https://github.com/iree-org/iree and https://github.com/nod-ai/iree-amd-aie) can independently update their lists of tests as their compiler and runtime code changes.

PR #205, opened 2024-04-25
Make torch MLIR build optional

kumardeepakamd:
Fixes https://github.com/nod-ai/SHARK-TestSuite/issues/49


e2eshark/requirements.txt
```diff
@@ -1,14 +1,16 @@
...
 optimum
-tabulate
+azure-storage-blob
+# install nightly build of torch_mlir 
+torch_mlir -f  https://github.com/llvm/torch-mlir-release/releases/expanded_assets/dev-wheels
 # install nightly build of iree-compiler and iree-runtime
 iree-compiler -f https://iree.dev/pip-release-links.html
 iree-runtime -f https://iree.dev/pip-release-links.html
-azure-storage-blob
-simplejson
```
saienduri, line 14: can you add a new line to this file

e2eshark/run.py
```diff
@@ -406,6 +385,110 @@ def runTorchMLIRGeneration(
...
         end = time.time()
         resultdict[curphase] = ["passed", end - start]
+    else:
+        iree_import_onnx = "iree-import-onnx"
+        curphase = phases[1]
+        logfilename = curphase + ".log"
+        if SHARED_IREE_BUILD:
+            # If local iree build provided use that instead of using the one installed
+            # in user's env
+            iree_import_onnx = SHARED_IREE_BUILD + "/tools/" + iree_import_onnx
```
saienduri, line 395: I don't think this will work. iree_import_onnx is not in the /tools area. it's a console_script python file: https://github.com/openxla/iree/blob/336ba1278c410b7dd46c2d01acbaa8ab29fc7b9a/compiler/setup.py#L453-L462. So we will have to do something like this f"{torch_mlir_pythonpath} python -m torch_mlir.tools.import_onnx " for iree_import_onnx

kumardeepakamd: So, I tested it and it works. The way it works is:
# install nightly build of iree-compiler and iree-runtime
iree-compiler -f https://iree.dev/pip-release-links.html
iree-runtime -f https://iree.dev/pip-release-links.html 
puts the iree-import-onnx tool in path. It is like an executable. 
If if SHARED_IREE_BUILD is provided, then the executable is in SHARED_IREE_BUILD + "/tools/" + iree_import_onnx else it is in path.

e2eshark/requirements.txt
kumardeepakamd: Just add an extra new line to make it easier to read?

e2eshark/run.py
kumardeepakamd: (e2e) <path>/iree-build/tools$ ls -1
build_config.txt
CMakeFiles
cmake_install.cmake
common
CTestTestfile.cmake
generate_embed_data
iree-benchmark-executable
iree-benchmark-module
iree-check-module
iree-compile
iree-convert-parameters
iree-cpuinfo
iree-create-parameters
iree-dump-instruments
iree-dump-module
iree-dump-parameters
iree-fatelf
iree-flatcc-cli
iree-lld
iree-mlir-lsp-server
iree-opt
iree-reduce
iree-run-mlir
iree-run-module
iree-tblgen
loader_test
test
testing
test-iree-compiler-api-test-binary
utils
It is a command line tool just like iree-compile and installed in env with pip install and in iree-build/tools when built.

saienduri: Yeah it is a CLI and is installed in env with pip install, but it is not in iree-build/tools unlike iree-compile as seen in your ls -1 command output right?

e2eshark/requirements.txt
saienduri: image
This red circle with a line means that we need a new line at the end of the file. Git is able to separate different files using this new line.

e2eshark/run.py
kumarxilinx: Makes sense. My bad. My eyes must have been tired. Good catch. So just pick from env always and not from local build? Or is there a way to reference it in local iree build?

e2eshark/requirements.txt
kumarxilinx: Got it. Will add.

kumardeepakamd: There seems to be new lines now

e2eshark/run.py
kumardeepakamd: Just push change to pick from env always. can incrementally add ability to pick from local iree build.

ScottTodd: > Or is there a way to reference it in local iree build?

It's there in the first comment:

> So we will have to do something like this f"{torch_mlir_pythonpath} python -m torch_mlir.tools.import_onnx " for iree_import_onnx

and there in the source file: https://github.com/iree-org/iree/blob/cc0c8266aa4d457705fcbb94804696ee0c8dcf18/compiler/bindings/python/iree/compiler/tools/import_onnx/__main__.py#L7-L16

`python -m iree.compiler.tools.import_onnx` is the general command when building the Python bindings from source (https://iree.dev/building-from-source/getting-started/#building-with-cmake).

You can also build the .whl and install from a local build instead of a distributed release, to get the same workflows in both places (I'd need to look up the specific command for that)

kumardeepakamd: I see. So, looks it like, a user has to explicitly configure CMake with IREE_BUILD_PYTHON_BINDINGS for iree-build/compiler/bindings/python to be created for ie2eshark run.py to be able to pick from PYTHONPATH. I am thinking that will be unnecessary burden on user. Given any user wanting to make change in import onnx code has to do in torch mlir code base anyway, it should be fine to pick iree-import-onnx from env always. For a local change in import_onnx pythonn package, user will have a local torch mlir build and use -c option to point to it.

saienduri (approved):
lgtm for now. thanks

PR #212, opened 2024-04-30
Add support to compare bin results

PhaneeshB:
- extends import_tests to use bins for input and output tensors for each test
- updates runtime xfails


e2eshark/_run_helper.py
ScottTodd, line 1: Can these changes to `e2eshark/` be moved to a different PR? (We also shouldn't generally be depending on code from one directory in the other, though I could see a case for sharing common utils)

iree_tests/onnx/import_tests_utils.py
```diff
@@ -0,0 +1,46 @@
...
+    np.dtype('uint8'): ('ui8', 'B') ,
+    np.dtype('float64'): ('f64', 'd') ,
+    np.dtype('float32'): ('f32', 'f') ,
+    np.dtype('float16'): ('f16', 'e') ,
+    np.dtype('bool'): ('i1', '?') ,
+}
+
+def getShapeString(torchtensor):
+    inputshape = list(torchtensor.shape)
+    inputshapestring = "x".join([str(item) for item in inputshape])
```
ScottTodd, line 22: follow PEP 8 naming? https://peps.python.org/pep-0008/#function-and-variable-names

`def get_shape_string(torch_tensor)`, `input_shape = ...`, etc.

Should at least be consistent within the utils file.

iree_tests/onnx/import_tests_utils.py
```diff
@@ -0,0 +1,46 @@
+import struct
+import numpy as np
```
ScottTodd, line 2: Please add a copyright header to source files

```python
# Copyright 2024 Advanced Micro Devices, Inc.
#
# Licensed under the Apache License v2.0 with LLVM Exceptions.
# See https://llvm.org/LICENSE.txt for license information.
# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```

iree_tests/onnx/import_tests_utils.py
```diff
@@ -0,0 +1,46 @@
...
+}
+
+def getShapeString(torchtensor):
+    inputshape = list(torchtensor.shape)
+    inputshapestring = "x".join([str(item) for item in inputshape])
+    dtype = torchtensor.dtype
+    if dtype in dtdict:
+        inputshapestring += f'x{dtdict[dtype][0]}'
+    else:
+        print("In getShapeString, found an unsupported data type", dtype)
```
ScottTodd, line 27: Could use `logging` here instead of `print`, or at least format the message:
`print(f"Warning: unsupported dtype in get_shape_string: '{dtype}'")`

iree_tests/onnx/import_tests_utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+def pack_np_arr(arr):
+    mylist = arr.flatten().tolist()
+    dtype = arr.dtype
+    # bf16 is converted to int16 when we `convert_io_proto` in `numpy_helper.to_array`
+    bytearr = b''
+    if dtype in dtdict:
+        bytearr = struct.pack(f"{len(mylist)}{dtdict[dtype][1]}", *mylist)
+    else:
+        print("In packTensor, found an unsupported data type", dtype)
```
ScottTodd, line 39: same comment about formatting log messages here

iree_tests/onnx/import_tests.py
```diff
@@ -131,8 +132,8 @@ def import_onnx_files(test_dir_path, imported_dir_path):
 
     # Convert input_*.pb and output_*.pb to .npy files.
```
ScottTodd, line 133: Please update the comment here (using binary .bin files instead of .npy files)

Could also add a note about endianness if we want to be really clear (the binary format won't work across systems with different endianness IIUC)

iree_tests/onnx/import_tests.py
```diff
@@ -141,15 +142,25 @@ def import_onnx_files(test_dir_path, imported_dir_path):
...
+        input_path_bin = (imported_dir_path / test_input.stem).with_suffix(".bin")
+        writeIOBinFile(t, input_path_bin)
+        test_data_flagfile_lines.append(f"--input={ss}=@{input_path_bin.name}\n")
     for i in range(len(test_outputs)):
         test_output = test_outputs[i]
         t = convert_io_proto(test_output, model.graph.output[i].type)
         if t is None:
             return False
         output_path = (imported_dir_path / test_output.stem).with_suffix(".npy")
         np.save(output_path, t)
```
ScottTodd, line 155: We can probably also delete the previous .npy files in this PR. I see both .npy and .bin files in https://github.com/nod-ai/SHARK-TestSuite/tree/iree_test_bin/iree_tests/onnx/node/generated/test_add_uint8, for example.

I think that's just a matter of dropping the `np.save` here and rerunning the importer, based on
```python

    # TODO(scotttodd): add flag to not clear output dir?
    print(f"Clearing old generated files from '{GENERATED_FILES_OUTPUT_ROOT}'")
    shutil.rmtree(GENERATED_FILES_OUTPUT_ROOT)
    GENERATED_FILES_OUTPUT_ROOT.mkdir(parents=True)
```

ScottTodd (commented):
Nice! Thanks for this.

iree_tests/onnx/import_tests.py
ScottTodd: Ah, I see this in the PR description

> keeps both npy and bin files for input and output for easy check of values using numpy

I guess that's okay. Could include a comment in the importer like this:
```python
np.save(output_path, t)  # Test using .bin but save .npy too for reference.
```

iree_tests/configs/config_onnx_cpu_llvm_sync.json
```diff
@@ -612,190 +612,76 @@
     "test_wrap_pad"
   ],
   "expected_run_failures": [
```
ScottTodd, line 614: To get the checks passing on this, https://github.com/nod-ai/SHARK-TestSuite/pull/211 should also be pulled in (can merge that then rebase/sync)

We're also considering trimming the config files in this repo and turning off "strict" XFAILs to make updating the files less of a chore. The general idea is that the config files _in this repo_ are for testing the test suite infrastructure itself, while the configs downstream _in repos like IREE_ are for showing which tests are continuously tested and passing/failing for some given source code and flags.

e2eshark/_run_helper.py
PhaneeshB: These changes started out with the intention for being moved to shared common utils but decided against introducing cross-directory util sharing. 
This at least brings forth the seemingly overall common utils for both directories and IMO should be just looked at as starting to lay the ground work for a possible root level shared utils structure in the future.

iree_tests/onnx/import_tests.py
PhaneeshB: That's right, I did come across the endianness aspect of using struct.packing, added the comment for big endian systems to import tests locally to have big endian bins. Thanks for noticing this!

iree_tests/onnx/import_tests_utils.py
PhaneeshB: can explore using logging project wide, updated the print stmt meanwhile

e2eshark/_run_helper.py
ScottTodd: The changes in that dir still look unrelated to the rest of the PR though? Keeping PRs focused makes them easier to review and improves commit log clarity.

ScottTodd (approved):

PR #214, opened 2024-05-01
[RFC] A design change proposal for the e2eshark tests

zjgarvey:
The broad goal of this design rework is to make it easier to add functionality and new tests to e2eshark. Towards this, there are sub-goals:

1. break up the complexity of run.py into a more coherent file structure.
2. remove the need to generate excessive boilerplate for model.py files, and generally aim to make adding tests as simple as possible. 
3. avoid running command-line scripts through python. This includes transitioning to using python bindings for torch-mlir and iree, and developing an import-friendly alternative to model.py files. 
4. avoid concatenating python files (again, by building a better import structure). 
5. minimize redundancy in the log (test-run) directories. For example, I believe the input and output tensors are actually stored three different ways (as a .bin, .pt, and .pkl), and this should not need to be the case. 

Abstractly there seem to be four input pieces to manage:

1. The model
2. The choice of frontend importer (e.g. iree-import-onnx, torch_mlir's fx importer, etc)
3. the choice of backend (e.g. compiling with iree from torch)
4. the input

Individual tests should just need to determine how to access or generate the model and how to generate a sample input.
Command line arguments should determine which frontend/backend combination is to be used during the test. 

The desired output pieces seem to be:

1. input/output/golden_output tensors
2. mlir/vmfb files used to generate output
3. stage and config-specific debugging info for tests which fail

### Current high-priority TODOs:

1. look into using pytest or another testing framework.
2. generate command-line reproducers for ease of debugging individual stages
3. come up with a better name for the directory than "alt_e2eshark".
4. identify functionality present in e2eshark that is currently lacking and should be migrated over. E.g. Chi's comment about the --report flag.
5. add more type hints to make it easier to track down definitions.
6. add (much) more to the README about adding tests and running them with various command line options.
7. look into removing torch-mlir as a python dependency entirely (perhaps optionally, so torch-mlir developers can still use the test suite for debugging op conversions).
8. Consider alternative options for storing cache directory paths, log directory paths, onnx model paths, etc. OnnxModelInfo is currently a bit clunky in needing to sometimes infer the log directory from the model path, etc.


alt_e2eshark/run.py
```diff
@@ -0,0 +1,52 @@
...
+       #TODO: Add logging/log_dir
+       inst = t.model_constructor(str(test_dir)) 
+       inputs = inst.construct_inputs()
+       golden_outputs = inst.forward(inputs)
+
+       mlir_module = config.mlir_import(inst)
+       buffer = config.compile(mlir_module)
+       callable_compiled_module = config.backend.load(buffer)
+       outputs = callable_compiled_module(inputs)
+       result = TestResult(name=t.unique_name, input=inputs,gold_output=golden_outputs,output=outputs)
```
ScottTodd, line 44: pytest could also help here. You could use a fixture or base test class to track this logic
https://docs.pytest.org/en/stable/explanation/fixtures.html

then could use https://pypi.org/project/pytest-xdist/ for multi-threading (or https://pypi.org/project/pytest-shard/ for running across machines)

ScottTodd (commented):
A mix of style nits (nice to establish before the code grows) and design questions/feedback.

alt_e2eshark/e2e_testing/backends.py
```diff
@@ -0,0 +1,41 @@
...
+from torch_mlir.passmanager import PassManager
+
+class SimpleIREEBackend(BackendBase):
+
+    def compile(self, module):
+        pipeline = "builtin.module(func.func(convert-torch-onnx-to-torch))"
+        with module.context as ctx:
+            pm = PassManager.parse(pipeline)
+            pm.run(module.operation)
+        return ic.tools.compile_str(str(module), input_type="torch", target_backends=["llvm-cpu"])
```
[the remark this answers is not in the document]
zjgarvey: > > avoid running command-line scripts through python. This includes transitioning to using python bindings for torch-mlir and iree, and developing an import-friendly alternative to model.py files.
> 
> A few thoughts here:
> 
>     * For iree_tests, running command-line scripts is an explicit design goal, so that developers working on the native tools can run (and debug) the tests directly (step by step) and the tests can be run on platforms without a python interpreter. The test runner there uses pytest and is thus written in Python, but other test runners could be written using CMake/CTest, Bazel, etc.

I see. It didn't make sense to me to run command line scripts from python if equivalent python bindings exist, since the bindings probably have a better api than organizing a command-line string. I suppose it is still possible with the class structure here to define a backend which uses command line scripting, if that is a design choice that we want to keep here. Either way I'd like to still dump all of the relevant files in an organized location, and have a log printing out equivalent command-line reproducers.

I guess if we wanted something python-agnostic, it might be best to upload the onnx files, and input.bin files, mlir to some storage that people could access if they wanted to run tests without python? 

> 
>     * For e2eshark, using Python bindings like here instead of native tools is a reasonable choice, though I'd keep a close eye on how to generate useful errors / logs / dumps. One of the first things that a compiler developer typically wants from a failing test case is a standalone reproducer.
> 
>     * In general, test suites should aim to use projects as users of those projects would. If the Python bindings prove difficult to use for whatever reason, that's something this sort of work could help highlight and drive improvements on.

Yeah, I'm finding the iree python bindings a bit difficult to wrap my head around right now, so when I get to know them a bit better I'll see if I have any suggestions to make.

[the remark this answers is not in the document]
zjgarvey: Yeah, good point. Would it be easy to add support for input_type="torch-onnx" at this point, for example? Or is it difficult for the AUTO option to detect the difference between torch-onnx and torch since they are using the same dialect?

alt_e2eshark/e2e_testing/backends.py
```diff
@@ -0,0 +1,41 @@
...
+    
+    def load(self, artifact):
+        config = rt.Config("local-task")
+        ctx = rt.SystemContext(config=config)
+        vm_module = rt.VmModule.copy_buffer(ctx.instance, artifact)
+        ctx.add_vm_module(vm_module)
+        for m in ctx.modules:
+            print(m)
```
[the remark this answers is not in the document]
zjgarvey: Yeah, I'd like to use logger when I add the actual logging functionality to each of these steps. 

This printing is simply because I don't know how to properly interface with the system context/bound modules yet. :(

alt_e2eshark/e2e_testing/backends.py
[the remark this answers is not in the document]
ScottTodd: > I guess if we wanted something python-agnostic, it might be best to upload the onnx files, and input.bin files, mlir to some storage that people could access if they wanted to run tests without python?

That's what iree_tests does (sort of, it starts from .mlir and has .onnx as a preprocessing step run offline). I've been thinking that model developers and people working on torch-mlir can use e2eshark as a hackable testing space then tests are exported from e2eshark into iree_tests. I have the start of a script to do such an export/import at https://github.com/nod-ai/SHARK-TestSuite/blob/main/iree_tests/pytorch/models/import_from_e2eshark.py. Having that sort of separation is also why I'm fine with making a different design choice here.

alt_e2eshark/e2e_testing/backends.py
```diff
@@ -0,0 +1,41 @@
...
+        def func(inputs):
+            return 0
+        return func
```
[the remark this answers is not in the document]
zjgarvey: I eventually want this method to return an callable which should represent the main function in the compiled module. I'm just not sure how to do it cleanly yet, so I'm going to look at some of the examples in, for example, SHARK's shark_inference.py.

alt_e2eshark/e2e_testing/framework.py
```diff
@@ -0,0 +1,83 @@
...
+from e2e_testing.storage import TestTensors
+from torch_mlir.ir import Module
+
+def generate_input_from_node(node: onnxruntime.capi.onnxruntime_pybind11_state.NodeArg):
+    if node.type == "tensor(float)":
+        return numpy.random.randn(*node.shape).astype(numpy.float32)
+    if node.type == "tensor(int)":
+        return numpy.random.randint(0, 10000, size=node.shape).astype(numpy.int32)
+    if node.type == "tensor(bool)":
+        return numpy.random.randint(0, 2, size=node.shape).astype(bool)
```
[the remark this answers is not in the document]
zjgarvey: Definitely, that's a good point.

alt_e2eshark/e2e_testing/backends.py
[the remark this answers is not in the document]
ScottTodd: Here is the current logic in IREE:
* https://github.com/iree-org/iree/blob/34449df33fbcc37d680abf37fdcfeae875ac04a1/compiler/plugins/input/Torch/PluginRegistration.cpp#L119-L125
    ```c++
        // ONNX is considered a superset of Torch. It runs all of the Torch
        // pipelines with an extra ONNX-specific preprocessing step.
        if (hasOnnx) {
          typeMnemonics.insert("onnx");
        } else if (hasTorch) {
          typeMnemonics.insert("torch");
        }
    ```
* https://github.com/iree-org/iree/blob/34449df33fbcc37d680abf37fdcfeae875ac04a1/compiler/plugins/input/Torch/PluginRegistration.cpp#L58-L69
    ```c++
        if (typeMnemonic == "onnx") {
          // ONNX input is a pre-processing step to torch.
          passManager.addNestedPass<func::FuncOp>(
              mlir::torch::onnx_c::createTorchOnnxToTorchPass());
        }
    
        if (typeMnemonic == "torch" || typeMnemonic == "onnx") {
          TorchInput::TorchToIREELoweringPipelineOptions torchOptions;
          torchOptions.strictSymbolicShapes = options.strictSymbolicShapes;
          TorchInput::createTorchToIREEPipeline(passManager, torchOptions);
          return true;
        }
    ```

We could change that to always run `createTorchOnnxToTorchPass` if any torch ops are found, or just keep the current behavior.

alt_e2eshark/e2e_testing/framework.py
```diff
@@ -0,0 +1,83 @@
...
+
+def get_sample_inputs_for_onnx_model(model_path):
+    s = onnxruntime.InferenceSession(model_path, None)
+    inputs = s.get_inputs()
+    sample_inputs = TestTensors(tuple([generate_input_from_node(node) for node in inputs]))
+    return sample_inputs 
+
+class OnnxModelInfo:
+    '''Stores information about an onnx test: the filepath to model.onnx, and how to construct/download it.'''
+    def __init__(self, onnx_model_path: str, dim_params: Optional[Dict[str, int]] = None):
```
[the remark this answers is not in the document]
zjgarvey: The biggest goal in reworking the structure is to have an easier time adding tests/functionality, so documentation will be really important. I'll make sure to flesh out the docstrings and add some readme files as I build this out more.

alt_e2eshark/e2e_testing/registry.py
```diff
@@ -0,0 +1,22 @@
...
+        raise Exception(
+            f"Duplicate test name: '{test_name}'. Please make sure that the function wrapped by `register_test` has a unique name."
+        )
+    _SEEN_NAMES.add(test_name)
+
+    # Store the test in the registry.
+    GLOBAL_TEST_LIST.append(
+        Test(
+            unique_name=test_name,
+            model_constructor=lambda path : test_class(path + "/" + test_name + ".onnx"),
```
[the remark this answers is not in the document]
zjgarvey: Ah, good point. I'm still figuring out how I want to manage this. I might just add the .onnx suffix in the class initializer instead (note: the OnnxModelInfo model attribute should represent a file path to the corresponding model.onnx file).

alt_e2eshark/e2e_testing/storage.py
```diff
@@ -0,0 +1,170 @@
+import io, torch, struct
+from typing import Tuple, Optional
+import numpy
+
+def loadTorchSave(filename):
```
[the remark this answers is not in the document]
zjgarvey: Yeah, this was stolen from the old run.py, so I'll try to update the style when migrating things in the future.

alt_e2eshark/onnx_tests/operators/model.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # Create the model (ModelProto)
+        onnx_model = make_model(graph)
+        onnx_model.opset_import[0].version = 19
+
+        onnx.save(onnx_model, self.model)
+
+register_test(AddModel, "add_test")
+
+class ConcatModel(OnnxModelInfo):
```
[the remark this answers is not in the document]
zjgarvey: I'd like to keep the number of model.py files as small as possible, so the import structure doesn't get out of hand. 

So similar to the torch-mlir e2e tests (from which some of this structure was motivated), I'd like to have a small number of thematically named test-suite files.

alt_e2eshark/run.py
```diff
@@ -0,0 +1,52 @@
+import sys
+from pathlib import Path
+sys.path.append(Path(__file__).parent)
+print(Path(__file__).parent)
```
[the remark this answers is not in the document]
zjgarvey: Haha, I'm trying to figure out how to best deal with the relative importing when using a venv. If you have a good read on this, let me know.

alt_e2eshark/run.py
zjgarvey: Yeah, my original goal was to keep the core functionality (besides trying to mininmize command-line scripting the running of concatenated python files from python) of the current run.py strategy, but simply factor each piece into a more appropriate python object. I'll definitely need to coordinate with someone about pytest and multi-threadded testing, since I know there are some subtleties there that I'm likely not prepared for yet.

zjgarvey:
@ScottTodd Thanks again for all of the very useful comments.

zjgarvey:
I'm finding this setup very helpful for testing the protected models. I think we should try to get this revived and added as a parallel directory so others can contribute to it.

zjgarvey:
@ScottTodd I've resolved most of the conversations so far. Some lingering concerns that I haven't fully addressed:

1. Using pytest as a testing framework. 
2. Path management for allowing imports: should I have the python file append to the user's path automatically, or have the user modify their PYTHONPATH before using?

I'd like to try to get this merged soon so that other people can contribute. Would you mind taking another look and seeing if this is in a position to get added?

AmosLewis:
```
python ./run.py --cachedir /proj/gdba/shark/cache -f onnx -g models -v
Stages to be run: ['setup', 'native_inference', 'mlir_import', 'torch_mlir', 'compilation', 'compiled_inference', 'post-processing']
Test list: ['DarkNet53', 'DarkNet53_vaiq', 'RAFT_vaiq_int8', 'pytorch-3dunet_vaiq_int8', 'FCN_vaiq_int8', 'u-net_brain_mri_vaiq_int8', 'resnet50_vaiq_int8', 'opt-125M-awq', 'opt-125m-gptq']
running test DarkNet53...
Begin download for e2eshark/onnx/models/DarkNet53/model.onnx.zip to /proj/gdba/shark/cacheDarkNet53/model.onnx.zip
Unzipping - /proj/gdba/shark/cacheDarkNet53/model.onnx.zip... 
Unzipping succeded. Look for extracted contents in /proj/gdba/shark/chi/src/SHARK-TestSuite/alt_e2eshark/test-run/DarkNet53/
        PASS
running test DarkNet53_vaiq...
Begin download for e2eshark/onnx/models/DarkNet53_vaiq/model.onnx.zip to /proj/gdba/shark/cacheDarkNet53_vaiq/model.onnx.zip
Unzipping - /proj/gdba/shark/cacheDarkNet53_vaiq/model.onnx.zip... 
Unzipping succeded. Look for extracted contents in /proj/gdba/shark/chi/src/SHARK-TestSuite/alt_e2eshark/test-run/DarkNet53_vaiq/
        FAILED (Numerics)
running test RAFT_vaiq_int8...
        FAILED (compilation)
running test pytorch-3dunet_vaiq_int8...
        FAILED (Numerics)
running test FCN_vaiq_int8...
        FAILED (Numerics)
running test u-net_brain_mri_vaiq_int8...
        FAILED (Numerics)
running test resnet50_vaiq_int8...
        PASS
running test opt-125M-awq...
Begin download for e2eshark/onnx/models/opt-125M-awq/model.onnx.zip to /proj/gdba/shark/cacheopt-125M-awq/model.onnx.zip
Unable to download model from public for opt-125M-awq.
Error - ResourceNotFoundError
Trying download from private storage
Please set AZ_PRIVATE_CONNECTION environment variable with connection string for private azure storage account
Unable to download model for opt-125M-awq.
Error - ValueError
Unzipping - /proj/gdba/shark/cacheopt-125M-awq/model.onnx.zip... 
Failed: path /proj/gdba/shark/cacheopt-125M-awq/model.onnx.zip does not exist!
        FAILED (native_inference)
running test opt-125m-gptq...
Using `disable_exllama` is deprecated and will be removed in version 4.37. Use `use_exllama` instead and specify the version with `exllama_config`.The value of `use_exllama` will be overwritten by `disable_exllama` passed in `GPTQConfig` or stored in your config file.
WARNING:auto_gptq.nn_modules.qlinear.qlinear_cuda:CUDA extension not installed.
WARNING:auto_gptq.nn_modules.qlinear.qlinear_cuda_old:CUDA extension not installed.
        FAILED (compilation)
Test Summary:
        PASSES: 2
        TOTAL: 9
results stored in /proj/gdba/shark/chi/src/SHARK-TestSuite/alt_e2eshark/test-run/
```
This is what I just run. The cachedir,  should add a / after it. otherwise in my case, it will download the model to /proj/gdba/shark/cacheDarkNet53_vaiq/model.onnx.zip not /proj/gdba/shark/cache/DarkNet53_vaiq/model.onnx.zip

And the CI need the --reports to generate report for all models daily to here https://github.com/nod-ai/e2eshark-reports .

renxida:
The functions in run.py could use some type hints. Would help me read this much faster.

alt_e2eshark/onnx_tests/models/model.py
```diff
@@ -0,0 +1,209 @@
...
+        opset_version = 21
+        super().__init__(name, onnx_model_path, cache_dir, opset_version)
+
+    def construct_model(self):
+        # try to find a .onnx file in the test-run dir
+        # if that fails, check for zip file in cache
+        # if that fails, try to download and setup from azure, then search again for a .onnx file
+
+        # TODO: make the zip file structure more uniform so we don't need to search for extracted files
+        model_dir = self.model.rstrip("model.nx")
```
renxida, line 28: Is this supposed to be model.onnx?

alt_e2eshark/e2e_testing/registry.py
```diff
@@ -0,0 +1,27 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+from e2e_testing.framework import Test
+
+GLOBAL_TEST_LIST = []
+
+_SEEN_NAMES = set()
+
+
+def register_test(test_class: type, test_name: str):
```
renxida, line 13: might be a good idea to add a 

```
def e2e_shark_test_case(test_name: str):
   return lambda test_class: register_test(test_class, test_name)
```

This will allow us to do
```
@e2e_shark_test_case("LLAMA3")
class Llama3TestCase:
  ...
```
with a decorator up front

renxida (approved):

renxida:
I love this and i think we should merge it. I can think of many improvements but most don't affect the interface.

alt_e2eshark/onnx_tests/models/model.py
```diff
@@ -0,0 +1,209 @@
...
+import torch
+from onnxruntime.tools.onnx_model_utils import make_dim_param_fixed, fix_output_shapes
+from e2e_testing import azutils
+from e2e_testing.framework import OnnxModelInfo
+from e2e_testing.registry import register_test
+from e2e_testing.storage import TestTensors
+from .protected_list import protected_models, models_with_postprocessing
+
+class AzureDownloadableModel(OnnxModelInfo):
+    def __init__(self, name: str, onnx_model_path: str, cache_dir: str):
```
renxida, line 18: I think we should stipulate this to always look for an onnx_model in the onnx_model path, and make cache_dir an environment variable instead. Reduce the number of paths to keep track of. Bonus points if we can even make the name a part of that.

renxida (approved):

alt_e2eshark/run.py
```diff
@@ -0,0 +1,415 @@
...
+    pipeline = REDUCE_TO_LINALG_PIPELINE if args.torchtolinalg else []
+    config = OnnxTestConfig(
+        str(TEST_DIR), SimpleIREEBackend(hal_target_backend=args.backend), pipeline
+    )
+
+    # get test list
+    test_list = get_tests(args.groups, args.test_filter)
+
+    # TODO: allow for no-run/mark xfails
+    # TODO: add better logging setup
```
ScottTodd, line 80: pytest would help here - no need to keep reinventing the wheel

alt_e2eshark/e2e_testing/framework.py
ScottTodd, line 1: This file is specific to onnx / torch. Should put that in the name/directory.

alt_e2eshark/onnx_tests/models/model.py
```diff
@@ -0,0 +1,209 @@
...
+        return TestTensors(processed_outputs)
+
+for t in protected_models:
+    register_test(AzureDownloadableModel, t)
+
+
+register_test(AzureDownloadableModel, "RAFT_vaiq_int8")
+register_test(AzureDownloadableModel, "pytorch-3dunet_vaiq_int8")
+register_test(AzureDownloadableModel, "FCN_vaiq_int8")
+register_test(AzureDownloadableModel, "u-net_brain_mri_vaiq_int8")
```
ScottTodd, line 73: Why multiple different models in a single test file? How about one model per file, with multiple test _cases_ in each file (e.g. different inputs or functions to the same program)?

alt_e2eshark/e2e_testing/backends.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+from iree import compiler as ireec
+from iree import runtime as ireert
+
+
+class SimpleIREEBackend(BackendBase):
+    '''This backend uses iree to compile and run MLIR modules for a specified hal_target_backend'''
+    def __init__(self, *, hal_target_backend="llvm-cpu"):
+        self.hal_target_backend = hal_target_backend
```
ScottTodd, line 34: Either the class name should be fully descriptive (`IREEBackendCPULocalTask`) or this should allow configuring both the compiler target backend (e.g. `llvm-cpu`) _and_ the runtime HAL device (e.g. `local-task`). For GPU backends like Vulkan, CUDA, and ROCm, the `ireert.Config("local-task")` call will use a different device like `ireert.Config("vulkan")`.

For naming and the set of configurations, see here for reference: https://github.com/iree-org/iree/tree/main/build_tools/pkgci/external_test_suite

alt_e2eshark/run.py
```diff
@@ -0,0 +1,415 @@
...
+    parser = argparse.ArgumentParser(prog="run.py", description=msg, epilog="")
+
+    # test config related arguments:
+    parser.add_argument(
+        "-b",
+        "--backend",
+        choices=["llvm-cpu", "amd-aie", "rocm", "vulkan"],
+        default="llvm-cpu",
+        help="Target backend i.e. hardware to run on",
+    )
```
ScottTodd, line 263: This should either include both
1. a compiler target backend 
2. a runtime HAL device

_or_ it should collect both of those into a single configuration

For examples of configurations, see
* https://github.com/iree-org/iree/tree/main/build_tools/pkgci/external_test_suite
* https://github.com/iree-org/iree/blob/main/build_tools/python/e2e_test_framework/definitions/iree_definitions.py
* https://github.com/iree-org/iree/blob/0934526e288315552650554da12d49b94845c7bf/integrations/tensorflow/python_projects/iree_tf/iree/tf/support/module_utils.py#L961-L990

alt_e2eshark/run.py
```diff
@@ -0,0 +1,415 @@
...
+    parser.add_argument(
+        "-f",
+        "--framework",
+        choices=["pytorch", "onnx", "tensorflow"],
+        default="onnx",
+        help="Run tests for given framework(s)",
+    )
```
ScottTodd, line 270: pytest would help here too - can choose to run tests under a specific directory or choose to run tests only matching a list of marks/labels

alt_e2eshark/e2e_testing/outputmod.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+import onnx
```
ScottTodd, line 6: This file is specific to onnx / torch. Should put that in the name/directory.

alt_e2eshark/e2e_testing/outputmod.py
```diff
@@ -0,0 +1,42 @@
...
+            print(nde)
+        counter += 1
+    if not node:
+        print("failed to find node")
+        import sys
+        sys.exit(1)
+    return node.output[0]
+
+model = load()
+modify_model(model, node_output_name(model, 2, "Conv"))
```
ScottTodd, line 42: What is this code doing? Any top level code (that runs at import time) is suspicious

ScottTodd (commented):
I wouldn't be opposed to landing and then continuing to iterate as a group. This does address a few of my critiques of the existing 'e2eshark' folder

alt_e2eshark/onnx_tests/models/model.py
zjgarvey: It should work either way. E.g.
```python
s = "abbc"
print(s.rstrip("cb")) # prints "a"
print(s.rstrip("bbc")) # also prints "a"
```
I'm fine changing it to "model.onnx" for readability.

alt_e2eshark/onnx_tests/models/model.py
zjgarvey: What would be in the file? 
```python
import stuff
register_test(AzureDownloadableModel, "modelname")
# end of file
```

I think it would be fine to split it up thematically like `azure_model.py`, `hf_vision_model.py` etc. But having potentially thousands of separate files just to add individual tests seemed like something I wanted to avoid. Is there a benefit to keeping the files separate for individual tests? E.g., does pytest require this structure?

The test-run directory that get generated when running tests does break up results/artifacts/logs into separate directories, if that is helpful context.

ScottTodd: If there are 1000s of test models each with 50 lines of code, I'd much rather have 1000s of test files than have a single 50000 line test file :P

Test runners/frameworks like pytest group by package, then file, then test case. They let you run all tests in a package, all tests in a file, a single test case, etc.

alt_e2eshark/e2e_testing/outputmod.py
zjgarvey: Ah, I need to work this out a bit more. This file can be used to modify an onnx model to return on like the 3rd conv op. I've been using it to try and debug numeric failures, but haven't settled on a plan for how to incorporate it into the test suite yet. Let me clean it up so at least it isn't running on import, and add a few comments on what these functions do.

alt_e2eshark/onnx_tests/models/model.py
zjgarvey: I don't think there is anything preventing this structure from having more files if they are getting too cluttered. The part that takes up the most space is definitely defining a new OnnxModelInfo child class, so it would make sense to separate those out into different files.

My feeling was that, since most models are available on azure for downloading directly, adding a new test would really be just specifying the name. I'd rather have:
```python
for t in some_list_of_testnames:
   register_test(AzureDownloadableModel, t)
```
with the list of testnames gotten from like a text file or json, than having thousands of files, or thousands of lines of any kind of code.

In any case, I can demo this change on the operator tests, since that model.py is already getting cluttered.

ScottTodd: I see your point about 
```python
for t in some_list_of_testnames:
   register_test(AzureDownloadableModel, t)
```
However, code like `Opt125MGPTQModelInfo` should go in its own file.

It's also nice being able to look at the directory tree and grep / file name search to find tests, but a monolithic `model.py` (should that be `models.py`?) needs its contents searched for a similar query. Collaboration like multiple people adding test cases is also easier with split files - lower risk of merge conflicts.

zjgarvey:
I'm going to update the PR name and initial comment with a TODO list before merging. Here is the original initial comment for reference. 

> This is a very rough sketch and some functions are not really implemented yet (for example, I'm not sure yet how to best handle the sample backend load method), but I'd like to put it up here for people to look at and give tips if they want to help out. 
> 
> The broad goal is to make it easier to add functionality and new tests to e2eshark. Towards this, there are sub-goals:
> 
> 1. break up the complexity of run.py into a more coherent file structure.
> 2. remove the need to generate excessive boilerplate for model.py files, and generally aim to make adding tests as simple as possible. 
> 3. avoid running command-line scripts through python. This includes transitioning to using python bindings for torch-mlir and iree, and developing an import-friendly alternative to model.py files. 
> 4. avoid concatenating python files (again, by building a better import structure). 
> 5. minimize redundancy in the log (test-run) directories. For example, I believe the input and output tensors are actually stored three different ways (as a .bin, .pt, and .pkl), and this should not need to be the case. 
> 
> Currently, I'm only setting up the configurations for onnx frontends, but I'm trying to do it in a way that would allow for a similar construction for torch.nn.Modules, for example. I'll add that structure soon. 
> 
> Abstractly there seem to be four input pieces to manage:
> 
> 1. The model
> 2. The choice of frontend importer (e.g. iree-import-onnx, torch_mlir's fx importer, etc)
> 3. the choice of backend (e.g. compiling with iree from torch)
> 4. the input
> 
> Individual tests should just need to determine how to access or generate the model and how to generate a sample input.
> Command line arguments should determine which frontend/backend combination is to be used during the test. 
> 
> The desired output pieces seem to be:
> 
> 1. input/output/golden_output tensors
> 2. mlir/vmfb files used to generate output
> 3. stage and config-specific debugging info for tests which fail
> 
> I'll continue adding to this draft over the next few days, so please feel free to ask questions or give advice.

alt_e2eshark/README.md
```diff
@@ -0,0 +1,111 @@
+ # e2eshark framework-to-iree-to-inference tests
```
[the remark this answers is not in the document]
zjgarvey: I'm adding this to the TODO list after merging.

alt_e2eshark/onnx_tests/models/model.py
zjgarvey: Adding this to the todo list after merging. Having to specify the cache directory each time is kind of annoying, and figuring out a better way to deal with path management in the class structure is also something we should look into.

PR #216, opened 2024-05-01
Update all vision model inputs and postprocess.

saienduri:
This commit changes the inputs from random to actual images for all vision models and also adds postprocessing pipeline to look at top k classifications for reference comparison. Results in 11 more models passing which were previously mismatching.


PhaneeshB:
Thanks for this! @saienduri

e2eshark/onnx/models/AlexNet_vaiq_int8/model.py
```diff
@@ -1,5 +1,11 @@
 import numpy, torch, sys
 import onnxruntime
+from PIL import Image
+import torchvision.transforms as transforms
+import requests
+
+def to_numpy(tensor):
```
PhaneeshB, line 7: NIT: since it's repeated in quite a few models we should probably move this to commonutils.py and import from there 
Otherwise LGTM!

saienduri: Good point, moved

kumardeepakamd (approved):
Good to get boiler plate code in tools utils.

PR #217, opened 2024-05-01
Update syntax of attention scripts for upstream change

qedawkins:


ScottTodd (approved):

qedawkins:
going to force merge here because there is a deadlock with making changes like this upstream: https://github.com/iree-org/iree/pull/17239

PR #220, opened 2024-05-02
Add support so we can have a central location for spec files

saienduri:
This commit adds support so we can use environment variables in our config files. This doesn't constrict us to only files in the directory. I moved the spec file out to a central location. The `list2command` is because when we use shell=true, 2nd and following arguments are for the shell itself, so we need to use a string to pass a command to the shell.

Update: PR has been updated to show helpful, easy to reproduce error messages:
Compile Failure Example: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8945969568
Runtime Failure Example: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/8946009337


iree_tests/conftest.py
```diff
@@ -375,12 +375,17 @@ def runtest(self):
             raise e
 
     def test_compile(self):
-        proc = subprocess.run(self.compile_args, capture_output=True, cwd=self.test_cwd)
+        compile_env = os.environ.copy()
+        compile_env["SPEC_LOCATION"] = os.getenv("SPEC_LOCATION", default=self.test_cwd)
```
ScottTodd, line 379: The name `SPEC_LOCATION` is specific to one type of file and is also not scoped to just IREE [tests]. If using an environment variable, I'm thinking we could build off of the `IREE_TEST_FILES` name with a name like `IREE_TEST_PATH_EXTENSION`.

Another idea would be to use substitutions like those in https://llvm.org/docs/CommandGuide/lit.html#substitutions. For example, we could look for occurrences of `%s` when loading the config .json file and substitute the config file's directory path in place of it. I think I like the idea of substitutions more than environment variables since that will involve less setup for CI jobs _and_ developers (no need for `SPEC_LOCATION: ${{ github.workspace }}/iree_tests/specs`). That will partly depend on what the reproducer CLI commands are - see my other comment.

iree_tests/pytorch/models/resnet50/attention_and_matmul_spec.mlir
ScottTodd, line 1: Glad to have these files deduplicated!

iree_tests/conftest.py
```diff
@@ -375,12 +375,17 @@ def runtest(self):
...
+        cmd = subprocess.list2cmdline(self.compile_args)
+        proc = subprocess.run(cmd, env=compile_env, shell=True, capture_output=True, cwd=self.test_cwd)
```
ScottTodd, line 381: We should be particularly careful about introducing "spooky action at a distance" (as Ben's words :P) here. When a test fails, the reproducer command that we log should be easy to replicate. Here's an example: https://github.com/iree-org/iree/actions/runs/8929912892/job/24559674205#step:9:909
`Invoked with: cd /home/esaimana/actions-runner/_work/iree/iree/SHARK-TestSuite/iree_tests/pytorch/models/sdxl-vae-decode-tank && iree-compile model.mlirbc --iree-hal-target-backends=vulkan-spirv -o model_gpu_vulkan_real_weights.vmfb`

What does the output look like with this new environment variable mapping?

I'm also thinking we should add the ability to log the compile and run commands _even for passing tests_. That way, we can point developers at the latest CI logs whenever they ask "what are the SDXL flags on main" and they'll be able to copy/paste from the location that ran the tests (even/especially if they passed). I think we can include logs for passing tests based on some changes in conftest.py and/or our flags/pytest.ini settings: https://docs.pytest.org/en/stable/how-to/logging.html.

iree_tests/conftest.py
saienduri: Ok we can do this too, although it seems more hardcoded for the position of these files. I guess we are not planning on changing the location of these files very often, but anytime we do, we would have to change the code and not a variable. Also, adding the substitution in the command in test_compile and test_run methods works too right?

iree_tests/conftest.py
saienduri: image
Lol interesting output. SPEC_LOCATION isn't expanded so I guess not the easiest to reproduce. also any idea what's going on with the run output?

ScottTodd: Ah! Glad we checked :P
The previous code expected a list, but `subprocess.list2cmdline` generates a string:
https://github.com/nod-ai/SHARK-TestSuite/blob/9e9774632d32674e9ffa44b0a2ffc054f6c590ad/iree_tests/conftest.py#L448-L449

I'm also thinking we could inline the `--flagfile=` in the output (maybe log both the flagfile version and the expanded version)

saienduri: Ahh okay got it! Yeah it would be good to have the flagfile arg expanded. I can expand the env variable for these error outputs too

iree_tests/conftest.py
saienduri: Is it ok to stick with environment variable because I'm also thinking if we want to expand this to the iree run flags. We would need to substitute the flagfile as well. But, we just add the --flagfile={location of flag file} right now, so not sure where substitution would happen (would have to happen in iree). Although we are adding another env variable, I think this might have the biggest scope.

iree_tests/conftest.py
saienduri: image

Thanks, looking like normal human readable output now :)

iree_tests/conftest.py
ScottTodd: I was imagining the substitution would happen here: https://github.com/nod-ai/SHARK-TestSuite/blob/9e9774632d32674e9ffa44b0a2ffc054f6c590ad/iree_tests/conftest.py#L94-L108

but it could also happen here: https://github.com/nod-ai/SHARK-TestSuite/blob/9e9774632d32674e9ffa44b0a2ffc054f6c590ad/iree_tests/conftest.py#L314-L320

I'm trying to focus on the reproducer and what a developer would author for a test, then work backwards from there.

In this case it is `cd /mnt/share/amdsharks/actions-runner/_work/SHARK-TestSuite/SHARK-TestSuite/iree_tests/pytorch/models/sdxl-scheduled-unet-3-tank && iree-run-module --module=model_gpu_rocm_real_weights.vmfb --device=hip --parameters=model=real_weights.irpa --module=sdxl_scheduled_unet_pipeline_fp16_rocm.vmfb --input=1x4x128x128xf16=@inference_input.0.bin --input=2x64x2048xf16=@inference_input.1.bin --input=2x1280xf16=@inference_input.2.bin --input=1xf16=@inference_input.3.bin --expected_output=1x4x128x128xf16=@inference_output.0.bin --expected_f16_threshold=0.8f --flagfile=real_weights_data_flags.txt`

The `cd [directory]` gets files like `@inference_input.1.bin` and `real_weights_data_flags.txt` on the PATH. For the compile command, `--iree-codegen-transform-dialect-library=${SPEC_LOCATION}/attention_and_matmul_spec.mlir` is missing. If we also log `set SPEC_LOCATION=/absolute/path/to/specs/` that should work. With a substitution, we'd instead have `--iree-codegen-transform-dialect-library=/absolute/path/to/specs/attention_and_matmul_spec.mlir`

saienduri:
Updated description with examples of compile and runtime failure messages now. Should be good to go

iree_tests/conftest.py
```diff
@@ -375,12 +375,19 @@ def runtest(self):
             raise e
 
     def test_compile(self):
-        proc = subprocess.run(self.compile_args, capture_output=True, cwd=self.test_cwd)
+        compile_env = os.environ.copy()
+        compile_env["IREE_TEST_PATH_EXTENSION"] = os.getenv("IREE_TEST_PATH_EXTENSION", default=self.test_cwd)
+        path_extension = compile_env["IREE_TEST_PATH_EXTENSION"]
+        cmd = subprocess.list2cmdline(self.compile_args)
+        cmd = cmd.replace("${IREE_TEST_PATH_EXTENSION}", f"{path_extension}")
+        proc = subprocess.run(cmd, env=compile_env, shell=True, capture_output=True, cwd=self.test_cwd)
```
ScottTodd, line 383: `IREE_TEST_PATH_EXTENSION` could be documented in the README at https://github.com/nod-ai/SHARK-TestSuite/tree/main/iree_tests and this code could have a comment, but that can be done later as we continue to refactor and make use of this

ScottTodd (approved):

iree_tests/conftest.py
saienduri: sounds good

PR #229, opened 2024-05-14
Fix home directory and relative path expansion due to cache_dir and cache_path variable confusion

renxida:


e2eshark/run.py
```diff
@@ -1307,28 +1307,26 @@ def main():
...
-    cache_path = os.path.abspath(cache_dir)
 
-    if not os.path.exists(cache_path):
-        print(f"ERROR: The Cache directory {cache_path} does not exist.")
+    cache_dir = os.path.expanduser(cache_dir)
+    cache_dir = os.path.abspath(cache_dir)
+
+    if not os.path.exists(cache_dir):
+        print(f"ERROR: The Cache directory {cache_dir} does not exist.")
         sys.exit(1)
```
ScottTodd, line 1317: Curious - why error and exit instead of creating the directory? Is the cache directory expected to be populated with some files at this point?

ScottTodd (approved):

e2eshark/run.py
renxida: i didn't do this part - - just moved it around a little. might put in another pr to fix this latr

PR #237, opened 2024-05-16
Added an onnx.SequenceConstruct test.

AWoloszyn:
This test is still failing in irre, as it requires a sequence as return to a function, which iree
does not seem to support at the moment.

### DIFF
--- a/e2eshark/onnx/operators/SequenceConstruct/model.py
+++ b/e2eshark/onnx/operators/SequenceConstruct/model.py
@@ -1,0 +1,78 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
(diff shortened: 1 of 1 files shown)


ScottTodd (approved):
Yeah this won't pass in IREE without a full look at the stack (and I'm not sure if we ever want to support sequences - would need to know if any programs we care about actually use them).

For the other ONNX tests, we don't even import from https://github.com/onnx/onnx/blob/main/onnx/backend/test/case/node/sequenceinsert.py or https://github.com/onnx/onnx/blob/main/onnx/backend/test/case/node/sequence_map.py:
https://github.com/nod-ai/SHARK-TestSuite/blob/757bc10677b433423b984181c6504af03ab10090/iree_tests/onnx/node/import_failures.txt#L32-L45

For that we would need to support other proto types here (but again, that's only useful if the underlying systems will ever support this): https://github.com/nod-ai/SHARK-TestSuite/blob/757bc10677b433423b984181c6504af03ab10090/iree_tests/onnx/import_tests.py#L43-L54

Fine to add more tests at the torch-mlir level but I'd proceed a bit carefully if intending to go further down into IREE with these operators.

PR #240, opened 2024-05-28
Import onnx tests with new iree-import-onnx on onnx 1.16.1.

ScottTodd:
This pulls in https://github.com/iree-org/iree/pull/17476 which now runs shape inference: `onnx.shape_inference.infer_shapes(model, data_prop=True)`.

Generated from an already configured venv with:
```
python -m pip install --upgrade --find-links https://iree.dev/pip-release-links.html iree-compiler
python -m pip install --upgrade onnx
python .\onnx\import_tests.py
```


ScottTodd:
Hmm still need to iterate on the CPU models xfails.

saienduri:
Yeah just need to update https://github.com/nod-ai/SHARK-TestSuite/blob/main/iree_tests/configs/config_sdxl_scheduled_unet_cpu_llvm_task.json

ScottTodd:
Thanks, that worked. When we refactor those jobs next, we could make those steps continue to show all errors instead of just the first. (I had a branch refactoring the jobs into a matrix, can pull that in...)

PR #241, opened 2024-05-28
Rework test_iree.yml to use a matrix strategy.

ScottTodd:
Small cleanup / deduplication here in this repo, mostly testing the workflow style to follow it in the IREE repo where we test across more configurations.


saienduri (approved):
Nice, this makes it cleaner for my different backend configs patch as well :)

PR #243, opened 2024-05-29
Add support for backend specific config files for sdxl models.

saienduri:
This commit adds support so that we can have different configuration files based on the backend for every sdxl model. This now gives us model and backend independence for all flags in sdxl

### DIFF
--- a/.github/workflows/test_iree.yml
+++ b/.github/workflows/test_iree.yml
@@ -101,11 +101,15 @@ jobs:
           - name: cpu_llvm_sync
             runs-on: nodai-amdgpu-w7900-x86-64
             models-config-file: config_pytorch_models_cpu_llvm_task.json
-            sdxl-config-file: config_sdxl_scheduled_unet_cpu_llvm_task.json
+            sdxl-prompt-encoder-config-file: config_sdxl_prompt_encoder_cpu_llvm_task.json
+            sdxl-unet-config-file: config_sdxl_scheduled_unet_cpu_llvm_task.json
+            sdxl-vae-decode-config-file: config_sdxl_vae_decode_cpu_llvm_task.json
           - name: gpu_mi250_rocm
             runs-on: nodai-amdgpu-mi250-x86-64
             models-config-file: config_gpu_rocm_models.json
-            sdxl-config-file: config_sdxl_scheduled_unet_gpu_rocm.json
+            sdxl-prompt-encoder-config-file: config_sdxl_prompt_encoder_gpu_rocm.json
+            sdxl-unet-config-file: config_sdxl_scheduled_unet_gpu_rocm.json
+            sdxl-vae-decode-config-file: config_sdxl_vae_decode_gpu_rocm.json
     env:
       VENV_DIR: ${{ github.workspace }}/.venv
       IREE_TEST_FILES: ~/iree_tests_cache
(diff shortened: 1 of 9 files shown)


.github/workflows/test_iree.yml
```diff
@@ -157,7 +161,23 @@ jobs:
...
+          source ${VENV_DIR}/bin/activate
+          pytest iree_tests/pytorch/models/sdxl-prompt-encoder-tank \
+            -rpfE \
+            -k real_weights \
+            --no-skip-tests-missing-files \
+            --capture=no \
+            --log-cli-level=info \
+            --timeout=1200 \
+            --durations=0 \
+            --config-files=${SDXL_PROMPT_ENCODER_CONFIG_FILE_PATH}
```
ScottTodd, line 177: this should be forwarded from the `matrix` through `env` like the others

```yml
    env:
      SDXL_CONFIG_FILE_PATH: iree_tests/configs/${{ matrix.sdxl-config-file }}
```

.github/workflows/test_iree.yml
```diff
@@ -101,11 +101,15 @@ jobs:
           - name: cpu_llvm_sync
             runs-on: nodai-amdgpu-w7900-x86-64
             models-config-file: config_pytorch_models_cpu_llvm_task.json
-            sdxl-config-file: config_sdxl_scheduled_unet_cpu_llvm_task.json
+            sdxl-prompt-encoder-config-file: config_sdxl_prompt_encoder_cpu_llvm_task.json
+            sdxl-unet-config-file: config_sdxl_scheduled_unet_cpu_llvm_task.json
+            sdxl-vae-decode-config-file: config_sdxl_vae_decode_cpu_llvm_task.json
```
ScottTodd, line 106: Fine for now, but let's figure out how to move these from the current `.json` config files to some other format (e.g. `.py` files or another `.json` format). I don't like special casing SDXL and this goes even further to special case _each submodel_.

.github/workflows/test_iree.yml
saienduri: Yeah just noticed that and updated, thanks

.github/workflows/test_iree.yml
saienduri: Ok that sounds good

.github/workflows/test_iree.yml
```diff
@@ -101,17 +101,23 @@ jobs:
           - name: cpu_llvm_sync
```
ScottTodd, line 101: I noticed this after merging, but I should have named this `cpu_llvm_task`. Could do that in a new PR or here.
```suggestion
          - name: cpu_llvm_task
```

.github/workflows/test_iree.yml
```diff
@@ -157,7 +163,23 @@ jobs:
             --durations=0 \
             --config-files=${MODELS_CONFIG_FILE_PATH}
 
-      - name: "Running real weights SDXL tests"
+      - name: "Running real weights SDXL prompt encoder tests"
+        id: prompt_encode
```
ScottTodd, line 167: `encode`, or `encoder`?
```suggestion
        id: prompt_encoder
```

.github/workflows/test_iree.yml
```diff
@@ -169,10 +191,28 @@ jobs:
...
+            --timeout=1200 \
+            --durations=0 \
+            --config-files=${SDXL_VAE_DECODE_CONFIG_FILE_PATH}
 
       - name: "Running SDXL ROCm pipeline benchmark"
-        if: contains(matrix.name, 'rocm')
+        if: |
+          contains(matrix.name, 'rocm') &&
+          steps.prompt_encode.outcome == 'success' && 
+          steps.unet.outcome == 'success' && steps.vae.outcome == 'success'
```
ScottTodd, line 215: Ah, good idea. I was wondering how to get the "continue if some tests fail but not these specific steps" behavior.

In particular, I don't want to run test jobs if repo checkout, venv setup, and other similar steps failed.

ScottTodd (approved):

.github/workflows/test_iree.yml
saienduri: encoder probably better

.github/workflows/test_iree.yml
saienduri: made change here

.github/workflows/test_iree.yml
saienduri: Yeah makes sense, we can add all the different job dependencies in another PR. Don't want to run the model testing if the azure download fails as well for example

Issue #105, opened 2024-03-11
"UnicodeEncodeError: 'charmap' codec can't encode character" running llama2-7b-GPTQ


ScottTodd:
I'm trying to get .mlir files from the test cases in [`e2eshark/pytorch/models/`](https://github.com/nod-ai/SHARK-TestSuite/tree/main/e2eshark/pytorch/models), so I can export them as standalone (no Python) tests ready for use in IREE (via the [`iree_tests/`](https://github.com/nod-ai/SHARK-TestSuite/tree/main/iree_tests) subproject.

I ran this on my Windows machine:
```
D:\dev\projects\SHARK-TestSuite\e2eshark
(.venv) λ python run.py --cachedir D:\dev\scratch\e2eshark \
  -c D:\dev\projects\torch-mlir\build \
  -i D:\dev\projects\iree-build \
  --tests pytorch/models/llama2-7b-GPTQ \
  --mode turbine
```

with this output (after 10 minutes running with my CPU at 100% and no intermediate output):
```
Cache Directory: D:\dev\scratch\e2eshark
Starting e2eshark tests. Using 4 processes
Torch MLIR build: D:\dev\projects\torch-mlir\build
IREE build: D:\dev\projects\iree-build
Test run directory: D:\dev\projects\SHARK-TestSuite\e2eshark\test-run
Since --tests or --testsfile was specified, --groups ignored
Framework:pytorch mode=turbine backend=llvm-cpu runfrom=model-run runupto=inference
Test list: ['pytorch/models/llama2-7b-GPTQ']
Test pytorch/models/llama2-7b-GPTQ failed [model-run]

Completed run of e2e shark tests

If using the upload feature, you can find a map of the test name to azure urls for your uploaded artifacts in upload_urls.json
```

model-run.log ends with:
```
Prompt: What is nature of our existence?
Response: Traceback (most recent call last):
  File "D:\dev\projects\SHARK-TestSuite\e2eshark\test-run\pytorch\models\llama2-7b-GPTQ\runmodel.py", line 46, in <module>
    print("Response:", tokenizer.decode(model_response[0]))
  File "C:\Program Files\Python311\Lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #247, opened 2024-05-30
Encode tokenizer outputs as utf8 before printing them.

ScottTodd:
Fixes https://github.com/nod-ai/SHARK-TestSuite/issues/105

These tokenizers sometimes output characters that cause trouble when printing with the default encoding on Windows, so explicitly encode as utf8. Not all tests currently generate such characters, but the extra safety seems helpful.

Sample response from whisper-small: `Response: b'<|startoftranscript|><|notimestamps|>What is nature of our existence?<|endoftext|>\xef\xbf\xbd\xef\xbf\xbd<|zh|>\xd0\xb7\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xe5\x84\x89\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd<|endoftext|>'`
* `UnicodeEncodeError: 'charmap' codec can't encode characters in position 82-83: character maps to <undefined>` (those are the `\xef\xbf` characters, unicode "replacement characters": https://stackoverflow.com/a/11162470, https://en.wikipedia.org/wiki/Specials_%28Unicode_block%29)

An alternate approach is to set the environment variable `PYTHONIOENCODING=utf-8` or `PYTHONUTF8=1`.

### DIFF
--- a/e2eshark/pytorch/models/bart-large/model.py
+++ b/e2eshark/pytorch/models/bart-large/model.py
@@ -43,7 +43,7 @@ model_response = model.generate(
     temperature=1.0,
 )
 print("Prompt:", prompt)
-print("Response:", tokenizer.decode(model_response[0]))
+print("Response:", tokenizer.decode(model_response[0]).encode("utf-8"))
 print("Input:", E2ESHARK_CHECK["input"])
 print("Output:", E2ESHARK_CHECK["output"])
 # For geneartive AI models, input is int and should be kept that way for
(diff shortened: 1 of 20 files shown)


saienduri (approved):
LGTM, thanks for making e2eshark windows friendly :)

PR #252, opened 2024-06-05
Update benchmarking to check benchmarking results and generate artifacts

saienduri:
This commit adds support, so that we actually make use of what we are benchmarking. We check against a golden time to make sure there are no regression for e2e and the individual submodels as well (unet, clip, vae). I also made the golden time a command line argument so that we can easily update it in the workflow file in either SHARK-TestSuite or iree whenever a patch has a performance boost, and there is no annoying repo dependence there.

There is also a job summary that is posted to the summary in CI, where there is an overview of the whole SDXL benchmarking that happened: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/9455228567


iree_tests/benchmarks/benchmark_sdxl_rocm.py
```diff
@@ -0,0 +1,63 @@
+import os
```
ScottTodd, line 1: Add copyright headers to source files

.github/workflows/test_iree.yml
```diff
@@ -215,4 +215,10 @@ jobs:
...
         run: |
           source ${VENV_DIR}/bin/activate
-          bash iree_tests/benchmarks/benchmark_sdxl_rocm.sh
+          pytest iree_tests/benchmarks/benchmark_sdxl_rocm.py --goldentime 5565
+    
+      - name: Upload SDXL ROCm benchmark artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: sdxl-benchmark-output
+          path: iree_tests/benchmarks/benchmark_out_rocm.txt
```
ScottTodd, line 224: > Also, I added an artifact that gets uploaded, so that the user can look through every iteration and the other data provided as well (median, std dev, cv, user counters, etc.) for the sdxl benchmark: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/9381843307/artifacts/1570376307

We can improve the utility here. This is uploading a human readable format to a file that still needs a few steps to view:
1. go to results page
2. download artifact
3. unzip
4. open text file

## Uploading artifacts

If uploading output, you could choose a different output format so automated systems can more easily process it: https://github.com/google/benchmark/blob/main/docs/user_guide.md#output-formats

Eventually I'd want benchmark results to be uploaded to a database. The in-tree benchmarking that IREE has uploads to https://perf.iree.dev/ but the underlying database (through https://github.com/google/dana) is not structured in a way that it can be easily queried.

## Posting comments

Another option is posting comments to the PR (like with IREE's [`post_benchmark_comment.yaml`](https://github.com/iree-org/iree/blob/main/.github/workflows/post_benchmark_comment.yaml)). That would skip all 4 steps above but I'd prefer for that to only happen if the PR author opted in or the results were particularly noteworthy. A PR touching Python bindings or updating documentation shouldn't get a comment about regressed benchmarks, for example.

## Job summaries

Some middle ground is a job summary:
* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary
* https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/

you'd generate some markdown that would appear on https://github.com/nod-ai/SHARK-TestSuite/actions/runs/9381843307/ , like https://github.com/iree-org/iree/actions/runs/9387594823#summary-25853165697 . That skips steps 2-4 above but still needs step 1.

## My recommendation

For now I'd suggest posting a job summary with human readable text (up to half of a screen or so? shouldn't be too noisy and should plan for more benchmark data being shown in the same location, such as from multiple machines or multiple models). You could keep the uploaded file either as the "console" format (default) or "json".

.github/workflows/test_iree.yml
```diff
@@ -215,4 +215,10 @@ jobs:
           steps.unet.outcome == 'success' && steps.vae.outcome == 'success'
         run: |
           source ${VENV_DIR}/bin/activate
-          bash iree_tests/benchmarks/benchmark_sdxl_rocm.sh
+          pytest iree_tests/benchmarks/benchmark_sdxl_rocm.py --goldentime 5565
```
ScottTodd, line 218: If you want print statements to appear in the output and plan on running through `pytest`, consider setting `--log-cli-level=info` here and then switching from `print()` to `logging.getLogger().info(...)`. I see that this is also swallowing the output for the benchmark: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/9381948995/job/25832310009?pr=252#step:11:20, making it _required_ to download the results. Ideally results would be streamed to the logs _and_ posted to a summary / uploaded as an artifact.

https://docs.pytest.org/en/stable/how-to/logging.html

For an example of how those fit together, see https://github.com/nod-ai/SHARK-TestSuite/pull/227

iree_tests/benchmarks/benchmark_sdxl_rocm.py
```diff
@@ -0,0 +1,63 @@
...
+from collections import namedtuple
+
+benchmark_dir = os.path.dirname(os.path.realpath(__file__))
+
+def launchCommand(scriptcommand):
+    print("Launching:", scriptcommand, "[ Proc:", os.getpid(), "]")
```
ScottTodd, line 7: I can't quite tell since this print didn't appear in logs (see my other comment), but I think I'd suggest this syntax be changed. I like to wrap some values in quotes when logging to make it obvious if the command is empty string and where the command stops and starts.

```suggestion
    print(f"Launching: '{scriptcommand}' [Proc: {os.getpid()}]")
```

iree_tests/benchmarks/benchmark_sdxl_rocm.py
```diff
@@ -0,0 +1,63 @@
...
+            errormsg,
+        )
+        return 1
+
+def run_sdxl_rocm_benchmark():
+    scriptcommand = (
+        "bash "
+        + f"{benchmark_dir}/benchmark_sdxl_rocm.sh"
+    )
+    return launchCommand(scriptcommand)
```
ScottTodd, line 25: It's a bit awkward to have a python script call a bash script when the python script could run all steps itself. See how https://github.com/iree-org/iree/blob/main/experimental/regression_suite/ireers/fixtures.py has a fixture for benchmarking

Windows also doesn't have the `bash` command by default, so this would not be portable to all systems in the way that a python script would.

iree_tests/benchmarks/benchmark_sdxl_rocm.py
```diff
@@ -0,0 +1,63 @@
...
+
+BenchmarkResult = namedtuple(
+    "BenchmarkResult", "benchmark_name time cpu_time iterations user_counters"
+)
+
+def decode_output(bench_lines):
+    benchmark_results = []
+    for line in bench_lines[3:]:
```
ScottTodd, line 33: Oh lol, yeah - if you're going to do this sort of parsing, just have the benchmark output a machine readable format instead :P

iree_tests/benchmarks/benchmark_sdxl_rocm.py
```diff
@@ -0,0 +1,63 @@
...
+                iterations=iterations,
+                user_counters=user_counters,
+            )
+        )
+    return benchmark_results
+
+def test_sdxl_rocm_benchmark(goldentime):
+    if run_sdxl_rocm_benchmark():
+        print("Running SDXL ROCm benchmark failed. Exiting")
+        return
```
ScottTodd, line 58: Control flow / error handling is confusing here. If run, failed?
Change the function name, check the return value, or throw an exception instead?

iree_tests/benchmarks/benchmark_sdxl_rocm.py
```diff
@@ -0,0 +1,63 @@
...
+    with open(f"{benchmark_dir}/benchmark_out_rocm.txt") as f:
+        bench_lines = f.readlines()
+    benchmark_results = decode_output(bench_lines)
+    benchmark_mean_time = int(benchmark_results[15].time.split()[0])
+    assert benchmark_mean_time < goldentime
```
ScottTodd, line 63: How does the output of this look when there is a regression? Might want to add some text https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

e.g.
`assert benchmark_mean_time < goldentime, "SDXL benchmark time should not regress"`

ScottTodd (changes requested):

.github/workflows/test_iree.yml
ScottTodd: Here's some example output from IREE's in-tree benchmarks: https://github.com/iree-org/iree/actions/runs/9381093834/job/25849050852#step:9:136

lots of json :P

iree_tests/benchmarks/conftest.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+def pytest_addoption(parser):
+    parser.addoption("--goldentime", action="store", default=5565, type=int, help="Golden time to test benchmark")
```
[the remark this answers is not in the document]
saienduri: 👍 also added in the backend

.github/workflows/test_iree.yml
saienduri: Did not know job summary was a thing. Thanks, seems like this is the best/cleanest fit for the benchmarking setup right now.

.github/workflows/test_iree.yml
saienduri: Switched to the log-cli-level flag and logger

iree_tests/benchmarks/benchmark_sdxl_rocm.py
saienduri: Switched to pure python implementation

iree_tests/benchmarks/benchmark_sdxl_rocm.py
saienduri: Checked return value and added comment

saienduri:
The w7900 runner (this is a different one from the iree repo's 2 runners) has gone down. Will get it back up, but this PR doesn't change any of the cpu testing code, so should be good.

iree_tests/benchmarks/benchmark_sdxl_rocm.py
```diff
@@ -1,36 +1,132 @@
...
-    scriptcommand = (
-        "bash "
-        + f"{benchmark_dir}/benchmark_sdxl_rocm.sh"
-    )
-    return launchCommand(scriptcommand)
+    exec_args = [
+        "iree-compile",
+        f"{benchmark_dir}/sdxl_pipeline_bench_f16.mlir",
+        "--iree-hal-target-backends=rocm",
+        "--iree-rocm-target-chip=gfx90a",
```
ScottTodd, line 35: Should this target chip be pulled from a CLI flag or environment variable? (The existing script hardcoded it too, but would be nice to allow running this on other GPUs)

iree_tests/benchmarks/benchmark_sdxl_rocm.py
```diff
@@ -1,36 +1,132 @@
...
+        "-o",
+        f"{benchmark_dir}/sdxl_full_pipeline_fp16_rocm.vmfb",
+    ]
+    # iree compile command for full sdxl pipeline
+    ret_value, stdout = run_iree_command(exec_args)
+    if ret_value == 1:
+        return 1, stdout
+    exec_args = [
+        "iree-benchmark-module",
+        "--device=hip://6",
```
ScottTodd, line 51: Similar question - let the device ID come from a flag or environment variable instead of hardcoding?

iree_tests/benchmarks/benchmark_sdxl_rocm.py
```diff
@@ -1,36 +1,132 @@
...
-    return launchCommand(scriptcommand)
+    exec_args = [
+        "iree-compile",
+        f"{benchmark_dir}/sdxl_pipeline_bench_f16.mlir",
+        "--iree-hal-target-backends=rocm",
+        "--iree-rocm-target-chip=gfx90a",
+        "--iree-global-opt-propagate-transposes=true",
+        "--iree-codegen-llvmgpu-use-vector-distribution",
+        "--iree-codegen-gpu-native-math-precision=true",
+        "--iree-rocm-waves-per-eu=2",
```
ScottTodd, line 39: If flags get shared between benchmarks and tests, we could move them to a flagfile. As a general rule of thumb for this sort of infrastructure, we should only benchmark models that we also test. If we benchmark a model without testing it too, then we might accidently end up benchmarking a model that doesn't actually perform the computations we want it to.

saienduri: Got it, the yaml file is setup so that the benchmarking job only executes if the testing for the models pass

ScottTodd: Are the flags the same? They should be pulled from the exact same file to avoid drift

iree_tests/benchmarks/conftest.py
```diff
@@ -1,8 +1,23 @@
...
+def goldentime_rocm_unet(request):
+    return request.config.getoption("--goldentime-rocm-unet-ms")
+
+@pytest.fixture
+def goldentime_rocm_clip(request):
+    return request.config.getoption("--goldentime-rocm-clip-ms")
+
+@pytest.fixture
+def goldentime_rocm_vae(request):
+    return request.config.getoption("--goldentime-rocm-vae-ms")
```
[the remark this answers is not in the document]
saienduri: At least from what I'm seeing online, the way to pass flags to a pytest program is through the conftest file. I didn't see a (easy) way to have the flags in the test file itself. There are some workarounds (https://stackoverflow.com/questions/59750480/pytest-and-argparse-code-works-in-use-but-tests-failing), but I just went with the pytest documentation for command line flags.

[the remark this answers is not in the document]
saienduri: Nice, the iree implementation is pretty cool. Thanks for sharing that. We will likely have to do something very similar with regex when we start to benchmark lot more models. I guess for now, with sdxl models and how much we care about every ms, this will do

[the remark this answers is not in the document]
ScottTodd: https://docs.python.org/3/library/os.html#os.environ environment variables would work, yeah?

iree_tests/benchmarks/benchmark_sdxl_rocm.py
saienduri: We can create a flag file in future for sure. Not seeing any major flags being shared amongst the benchmarks at the moment

.github/workflows/test_iree.yml
```diff
@@ -215,10 +215,12 @@ jobs:
...
-      - name: Upload SDXL ROCm benchmark artifacts
-        uses: actions/upload-artifact@v4
-        with:
-          name: sdxl-benchmark-output
-          path: iree_tests/benchmarks/benchmark_out_rocm.txt
+          pytest iree_tests/benchmarks/benchmark_sdxl_rocm.py \
+            --goldentime-rocm-e2e-ms 1661.2 \
+            --goldentime-rocm-unet-ms 450.4 \
+            --goldentime-rocm-clip-ms 18.9 \
+            --goldentime-rocm-vae-ms 288.5 \
```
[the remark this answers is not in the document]
saienduri: These numbers are pretty stable. I've tried around 15 times on the mi250 gpus. It failed once. I also added --reruns 3 to the pytest. So, it should be very unlikely that we get a false fail. Just want to be as strict as possible

iree_tests/benchmarks/benchmark_sdxl_rocm.py
```diff
@@ -52,12 +148,53 @@ def decode_output(bench_lines):
...
+def test_sdxl_rocm_benchmark(goldentime_rocm_e2e, goldentime_rocm_unet, 
+    goldentime_rocm_clip, goldentime_rocm_vae):
+    # if the benchmark returns 1, benchmark failed
+    job_summary_lines = []
+
+    # e2e benchmark
+    ret_value, output = run_sdxl_rocm_benchmark()
+    benchmark_mean_time = job_summary_process(ret_value, output)
+    assert benchmark_mean_time < goldentime_rocm_e2e, "SDXL e2e benchmark time should not regress"
+    mean_line = f"E2E Benchmark Time: {str(benchmark_mean_time)}"
```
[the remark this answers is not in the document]
saienduri: Good idea, made the update.

iree_tests/benchmarks/benchmark_sdxl_rocm.py
saienduri: Yeah that makes more sense, did CLI flag

iree_tests/benchmarks/benchmark_sdxl_rocm.py
saienduri: Oh your comments didn't load for some reason. Yes, the flags are the same as the ones in the iree_tests/configs/ json files. It seems a bit messy to copy the process to convert and check json files that you did originally here too. These are just the input and ouput dimensions, so pretty rare that they will change. What do you think?

iree_tests/benchmarks/conftest.py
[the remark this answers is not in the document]
saienduri: Yup environment variables would work. I feel like we already have a good amount of them in the yml file. As a user, I prefer command line flags lol (also just want people to just have to edit that one line in the yml file whenever the golden values are changed with perf boosts). But, maybe that's just me. Fine with either

.github/workflows/test_iree.yml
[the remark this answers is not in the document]
ScottTodd: /me puts on his statistics hat

At least within a single run we can reference the measured standard deviation to determine how much precision is warranted.

On this run: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/9457184418/job/26050493249?pr=252#step:11:35
* e2e real_time_mean 1659ms
* e2e real_time_stddev 0.569ms
* e2e iterations 3 (not likely enough to be statistically significant?)

For this 1661.2 value, we're saying that we know the precision down to the nearest 0.1ms, yet the standard deviation is near 0.5ms. If we assume a normal distribution, 95% of values would fall within two standard deviations (~1.1ms) of the mean

I'm also not sure how much we can trust the distribution to be normal and consistent, across machines, reboots, other environment settings (other programs running on the system), etc.

---

Same exercise for the other benchmarks:

* unet real_time_mean 449ms
* unet real_time_stddev 0.151ms
* golden time 450.4

Maybe round up to 450.5ms?

* clip (prompt encoder) real_time_mean 18.7ms
* clip real_time_stddev 0.155ms
* golden time 18.9ms

Maybe round up to 19.0ms?

* vae real_time_mean 287ms
* vae real_time_stddev 0.103ms
* golden time 288.5ms

Fine to keep at current value

---

> I also added --retries 3 to the pytest. So, it should be very unlikely that we get a false fail.

Interesting... https://github.com/nod-ai/SHARK-TestSuite/actions/runs/9457184418/job/26050493249?pr=252#step:11:62 had a retry on it. I guess that's okay and we can monitor how it works in practice.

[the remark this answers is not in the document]
saienduri: Thanks for doing the statistical analysis and giving me confidence from the mathematical side :). Rounded as suggested, so we don't give off that we know the precision down to the nearest 0.1ms

iree_tests/benchmarks/benchmark_sdxl_rocm.py
ScottTodd: What I had in mind was extracting these sorts of flags into a flagfile.txt: https://github.com/nod-ai/SHARK-TestSuite/blob/16812962da23c85302c03ef505f4c654cb54b46c/iree_tests/configs/config_sdxl_scheduled_unet_gpu_rocm.json#L8-L18

Then in that file using `--flagfile=${IREE_TEST_PATH_EXTENSION}/sdxl_rocm_compile_flags.txt` and similarly sourcing the file here too.

Juggling this many flags across configurations gets unwieldy no matter how we do it though, so we should be moving towards configuration in .mlir files or some other purpose-built API IMO. I guess for now let's keep the duplication and just be careful about letting the complexity spread too much.

ScottTodd (approved):
Looks good to land. A few things to keep in mind as this gets used more:

* How to scale this to multiple models, multiple test machines, multiple backends (CPU/GPU/etc.)
* How to manage the golden times and tolerances as the performance changes or the environment is less stable than we'd like
* How visible these metrics are on PRs
* If the mi250 results are a good enough proxy for mi300 (or whatever else we want to be measuring on 3+ months from now)

iree_tests/benchmarks/benchmark_sdxl_rocm.py
saienduri: Ah got it. Yeah we can definitely do that

PR #254, opened 2024-06-06
onnx `model.py` fix

IanWood1:
https://github.com/nod-ai/SHARK-Turbine/issues/683
When using iree with local changes that fixed `iree-compile` compilation errors, the test started to fail in inference (`iree-run-module`). I was specifically running the RAFT test. This seemed to be due to accessing `E2ESHARK_CHECK` with "inputs" instead of "input".
After fixing that I got a different inference error due to a mismatch between expected outputs and the outputs of the vmfb. I grepped  `OUT_OF_RANGE` and found other tests failing in the same way (found in model's `inference.log` file). There could be other tests that are failing earlier than inference that have this same issue.


AmosLewis (approved):
LGTM

PR #265, opened 2024-06-14
Update benchmarking step to also test compilation stats

saienduri:
This commit adds support so that the benchmarking step also tests the compilation statistics of the model. I moved all the benchmarking into a sdxl subdirectory of benchmarking, so we can have different conftest.py for every model. This way we don't flood one conftest file with all the parameters needed for all the models.

Instead of <=, do you think I should maybe go with the == for the asserts regarding dispatches and size? (It will force contributors to update when it changes, and it is a static value)

Here is what updated job summary looks like (https://github.com/nod-ai/SHARK-TestSuite/actions/runs/9586667587):
image


iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py
```diff
@@ -181,6 +188,27 @@ def test_sdxl_rocm_benchmark(goldentime_rocm_e2e, goldentime_rocm_unet,
     job_summary_lines.append(mean_line)
     logging.getLogger().info(mean_line)
 
+    # unet compilation stats check
+    with open(f"{scheduled_unet_dir}/compilation_info.json", 'r') as file:
+        comp_stats = json.load(file)
+    unet_dispatch_count = int(comp_stats["stream-aggregate"]["execution"]["dispatch-count"])
+    assert unet_dispatch_count <= goldendispatch_rocm_unet, "SDXL scheduled unet dispatch count should not regress"
```
ScottTodd, line 195: It looks like these `assert`s will immediately fail the test. What we probably want is for the summary to always be generated and for _all_ errors to be reported, not just the first error. A way of doing that under pytest is using https://github.com/okken/pytest-check

1. add `pytest-check` to a `requirements.txt` file
2. add `from pytest_check import check`
3. modify these `assert` calls to use `check`, e.g.

    ```python
    check.less_equal(unet_dispatch_count, goldendispatch_rocm_unet, "SDXL scheduled unet dispatch count should not regress")
    ```
```

File: iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py [hunk 2]
```diff
@@ -172,6 +176,9 @@ def test_sdxl_rocm_benchmark(goldentime_rocm_e2e, goldentime_rocm_unet,
     job_summary_lines.append(mean_line)
     logging.getLogger().info(mean_line)
 
+    # new line
+    job_summary_lines.append("\n")
```
ScottTodd, line 180: drop trivial comment (or add more context like "New line separator between result sections.")
```suggestion
    job_summary_lines.append("\n")
```

iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py
```diff
@@ -181,6 +188,27 @@ def test_sdxl_rocm_benchmark(goldentime_rocm_e2e, goldentime_rocm_unet,
     job_summary_lines.append(mean_line)
     logging.getLogger().info(mean_line)
 
+    # unet compilation stats check
+    with open(f"{scheduled_unet_dir}/compilation_info.json", 'r') as file:
+        comp_stats = json.load(file)
+    unet_dispatch_count = int(comp_stats["stream-aggregate"]["execution"]["dispatch-count"])
+    assert unet_dispatch_count <= goldendispatch_rocm_unet, "SDXL scheduled unet dispatch count should not regress"
+    compilation_line = (f"Scheduled Unet Dispatch Count: {unet_dispatch_count}"
+                        f" (golden dispatch count {goldendispatch_rocm_unet})")
```
ScottTodd, line 197: You could format these results into a table

Current text:
```
SDXL Benchmark Summary:
E2E Benchmark Time: 1659.0 ms (golden time 1661.5 ms)

Scheduled Unet Benchmark Time: 449.0 ms (golden time 450.5 ms)
Scheduled Unet Dispatch Count: 1718 (golden dispatch count 1718)
Scheduled Unet Binary Size: 2088217 bytes (golden binary size 2088217 bytes)

Prompt Encoder Benchmark Time: 16.7 ms (golden time 19.0 ms)
Prompt Encoder Dispatch Count: 1571 (golden dispatch count 1571)
Prompt Encoder Binary Size: 785493 bytes (golden binary size 785493 bytes)

VAE Decode Benchmark Time: 288.0 ms (golden time 288.5 ms)
VAE Decode Dispatch Count: 250 (golden dispatch count 250)
VAE Decode Binary Size: 762067 bytes (golden binary size 762067 bytes)
```

Markdown table (using empty rows as dividers, could leave those off):

Metric | Current value | Expected/golden value
-- | -- | --
E2E Benchmark Time | 1659.0 ms |1661.5 ms
 | | 
Scheduled Unet Benchmark Time | 449.0 ms | 450.5 ms
Scheduled Unet Dispatch Count | 1718 | 1718
Scheduled Unet Binary Size | 2088217 bytes | 2088217 bytes
 | | 
Prompt Encoder Benchmark Time | 16.7 ms | time 19.0 ms
Prompt Encoder Dispatch Count | 1571 | 1571
Prompt Encoder Binary Size | 785493 bytes | 785493 bytes
 | | 
VAE Decode Benchmark Time | 288.0 ms | time 288.5 ms
VAE Decode Dispatch Count | 250 | 250
VAE Decode Binary Size | 762067 bytes | 762067 bytes

Markdown source:
```

Metric | Current value | Expected/golden value
-- | -- | --
E2E Benchmark Time | 1659.0 ms |1661.5 ms
 | | 
Scheduled Unet Benchmark Time | 449.0 ms | 450.5 ms
Scheduled Unet Dispatch Count | 1718 | 1718
Scheduled Unet Binary Size | 2088217 bytes | 2088217 bytes
 | | 
Prompt Encoder Benchmark Time | 16.7 ms | time 19.0 ms
Prompt Encoder Dispatch Count | 1571 | 1571
Prompt Encoder Binary Size | 785493 bytes | 785493 bytes
 | | 
VAE Decode Benchmark Time | 288.0 ms | time 288.5 ms
VAE Decode Dispatch Count | 250 | 250
VAE Decode Binary Size | 762067 bytes | 762067 bytes
```

---

Or could sort by metric:

Benchmark | Current time (ms) | Expected/golden time (ms)
-- | -- | --
E2E† | 1659.0 |1661.5
Scheduled Unet | 449.0 | 450.5
Prompt Encoder | 16.7 | 19.0
VAE Decode | 288.0 | 288.5

† E2E = Encode + Scheduled Unet * 3 + Decode

Benchmark | Current dispatch count | Expected/golden dispatch count
-- | -- | --
Scheduled Unet | 1718 | 1718
Prompt Encoder | 1571 | 1571
VAE Decode | 250 | 250


Benchmark | Current binary size (bytes) | Expected/golden binary size (bytes)
-- | -- | --
Scheduled Unet | 2088217 | 2088217
Prompt Encoder | 785493 | 785493
VAE Decode | 762067 | 762067

markdown source:
```
Benchmark | Current time (ms) | Expected/golden time (ms)
-- | -- | --
E2E† | 1659.0 |1661.5
Scheduled Unet | 449.0 | 450.5
Prompt Encoder | 16.7 | 19.0
VAE Decode | 288.0 | 288.5

† E2E = Encode + Scheduled Unet * 3 + Decode

Benchmark | Current dispatch count | Expected/golden dispatch count
-- | -- | --
Scheduled Unet | 1718 | 1718
Prompt Encoder | 1571 | 1571
VAE Decode | 250 | 250


Benchmark | Current binary size (bytes) | Expected/golden binary size (bytes)
-- | -- | --
Scheduled Unet | 2088217 | 2088217
Prompt Encoder | 785493 | 785493
VAE Decode | 762067 | 762067
```

I think I like the last version there

iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py
```diff
@@ -181,6 +188,27 @@ def test_sdxl_rocm_benchmark(goldentime_rocm_e2e, goldentime_rocm_unet,
...
+    module_path = f"{scheduled_unet_dir}/model_gpu_rocm_real_weights.vmfb"
+    binary_size = Path(module_path).stat().st_size
+    assert binary_size <= goldensize_rocm_unet, "SDXL scheduled unet binary size should not get bigger"
+    compilation_line = (f"Scheduled Unet Binary Size: {binary_size} bytes"
+                        f" (golden binary size {goldensize_rocm_unet} bytes)")
+    job_summary_lines.append(compilation_line)
+    logging.getLogger().info(compilation_line)
+
+    # new line
+    job_summary_lines.append("\n")
```
ScottTodd, line 210: Same here - drop trivial comments (or expand on the intent)

iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py
saienduri: Cool, updated

iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py
saienduri: This is a lot nicer, made the updates for markdown :)

iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py
```diff
@@ -159,46 +163,145 @@ def job_summary_process(ret_value, output):
...
                  f" (golden time {goldentime_rocm_unet} ms)")
-    job_summary_lines.append(mean_line)
+    unet_mean_time = benchmark_mean_time
     logging.getLogger().info(mean_line)
 
+    # unet compilation stats check
+    with open(f"{scheduled_unet_dir}/compilation_info.json", 'r') as file:
+        comp_stats = json.load(file)
+    unet_dispatch_count = int(comp_stats["stream-aggregate"]["execution"]["dispatch-count"])
+    check.less_equal(unet_dispatch_count, goldendispatch_rocm_unet, "SDXL scheduled unet dispatch count should not regress")
```
ScottTodd, line 192: Dispatch count can change and reducing it may improve performance (e.g. enabling more fusions will reduce dispatch count). @MaheshRavishankar what behavior do you want here? 
A) check that the value doesn't change at all, so every PR that affects the dispatch count will need to set these to new values
B) check that the value doesn't change beyond some threshold (e.g. 20%)
C) just log the value without failing the test if it changes from some known value

```
            --goldendispatch-rocm-unet 1718 \
            --goldendispatch-rocm-clip 1571 \
            --goldendispatch-rocm-vae 250 \
```

iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py
```diff
@@ -159,46 +163,145 @@ def job_summary_process(ret_value, output):
...
+    compilation_line = (f"Scheduled Unet Dispatch Count: {unet_dispatch_count}"
+                        f" (golden dispatch count {goldendispatch_rocm_unet})")
+    logging.getLogger().info(compilation_line)
+
+    module_path = f"{scheduled_unet_dir}/model_gpu_rocm_real_weights.vmfb"
+    binary_size = Path(module_path).stat().st_size
+    check.less_equal(binary_size, goldensize_rocm_unet, "SDXL scheduled unet binary size should not get bigger")
```
ScottTodd, line 199: Similar comment here - increasing binary size can improve performance (e.g. if including more specialized kernels, or generating more unrolled loops). At the same time, decreasing binary size can improve performance too (e.g. folding sequences of ops or reducing the number of barriers in a program)

@MaheshRavishankar should this just check the exact value? Or log the value but not check it against a reference?

iree_tests/benchmarks/sdxl/conftest.py
```diff
@@ -0,0 +1,63 @@
+import pytest
+
+def pytest_addoption(parser):
+    parser.addoption("--goldentime-rocm-e2e-ms", action="store", default=1661.5, type=float, help="Golden time to test benchmark")
+    parser.addoption("--goldentime-rocm-unet-ms", action="store", default=450.5, type=float, help="Golden time to test benchmark")
```
ScottTodd, line 5: I think these defaults should be removed and the flag options should be required by all uses of the script. Otherwise, this encodes a set of values at this point in time for one GPU.

iree_tests/benchmarks/sdxl/conftest.py
```diff
@@ -0,0 +1,63 @@
...
+    parser.addoption("--goldentime-rocm-vae-ms", action="store", default=288.5, type=float, help="Golden time to test benchmark")
+    parser.addoption("--goldendispatch-rocm-unet", action="store", default=1718, type=int, help="Golden dispatch count to test benchmark")
+    parser.addoption("--goldendispatch-rocm-clip", action="store", default=1571, type=int, help="Golden dispatch count to test benchmark")
+    parser.addoption("--goldendispatch-rocm-vae", action="store", default=250, type=int, help="Golden dispatch count to test benchmark")
+    parser.addoption("--goldensize-rocm-unet-bytes", action="store", default=2088217, type=int, help="Golden vmfb size to test benchmark")
+    parser.addoption("--goldensize-rocm-clip-bytes", action="store", default=785493, type=int, help="Golden vmfb size to test benchmark")
+    parser.addoption("--goldensize-rocm-vae-bytes", action="store", default=762067, type=int, help="Golden vmfb size to test benchmark")
+    parser.addoption("--gpu-number", action="store", default=6, type=int, help="IREE GPU device number to test on")
```
ScottTodd, line 14: I'd default this to 0 instead of 6.

saienduri: Others also use this node. I think 6 is best bet to not have any interference

ScottTodd: Yeah - using node 6 on the current test machine is all fine, but I don't think that should be baked into the script that gets used across repositories as a default.

iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py
MaheshRavishankar: I think what would work is dispatch count shouldnt increase, but it can decrease. I think doing (A) is fine if it is easy to change that value.

saienduri: Sounds good, easy to update, so let's go with checking that the value doesn't change at all to catch any changes in the compilation statistics.

iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py
saienduri: The checks are not allowing the retries to occur. The retries are sometimes necessary to get the job passing. With it, we are able to be pretty strict with the values and have pretty solid testing so far. I moved the asserts to the end after the report is generated, so that we can stick with what's working for now (asserts/retries) and still get the report to show the status of every statistic.

ScottTodd: Ehh, this still only shows the first assert though.

```python
    assert benchmark_e2e_mean_time <= goldentime_rocm_e2e, "SDXL e2e benchmark time should not regress"
    assert benchmark_unet_mean_time <= goldentime_rocm_unet, "SDXL unet benchmark time should not regress"
    assert unet_dispatch_count == goldendispatch_rocm_unet, "SDXL scheduled unet dispatch count should not regress"
```

Do you have any theories on why the retries are helping? Are there other workloads running on the system? You could also try running with `--benchmark_min_warmup_time` (scaled to get enough iterations in).

I'd really like to avoid retries since they can complicate logs and interact poorly with non-flaky failures.

Otherwise I guess we could proceed with this and iterate on it later.

ScottTodd: I'm also concerned looking at those asserts that a developer would see "SDXL scheduled unet binary size should not get bigger" and then only update the `unet` binary size when the `vae decode` binary size might also need updating.

ScottTodd (approved):

iree_tests/benchmarks/sdxl/benchmark_sdxl_rocm.py
saienduri: Yeah, looks like --benchmark_min_warmup_time and increasing iterations to 10 works from my testing. One thing to note, if you rerun a job and the job recompiles the vmfb, it may slightly change the binary size. But, the timings always passed when I tested, so I will go with this for now :)

PR #271, opened 2024-06-25
Support for new test suite for special models.

saienduri:
This commit adds the support for a new test suite for special models where we can use custom flags, tolerances, and configurations based on the backend/model compared to iree_tests where we stick with the most basic testing configurations.
In this setup, we don't have to deal with any config files for iree flags.
I also generalized the tools being used in this new test suite, and changed the old test suite to use these generalized tools.
This way we have uniform testing (can expect same behavior from all suites). To do this I created a python package for iree_special_models which is also now pip installed in the workflow file.
The conftest for iree_tests is simple again and has no weird environment variable expansions

Each model test usually has 3 files (scheduled unet has an extra mlir to test the scheduling (don't need those vmfbs anymore)):
1) source mlir
2) json file for what remote files to download
3) python testing file

More info can be found in the README.md for this directory


iree_tests/conftest.py
```diff
@@ -13,6 +13,7 @@
 import os
 import pytest
 import subprocess
+from ireers import *
```
ScottTodd, line 16: How about putting `ireers` in a `common/` folder?

Current (reaching deep into a sibling):
```
iree_special_models/
  ireers/
  sd3/
    (depends on ../ireers/)
iree_tests
  (depends on ../iree_special_models/ireers)
```

Other options:

*
    ```bash
    # This puts "common" at the root, possibly shared with e2eshark
    common/
      fixtures.py
    iree_special_models/
    iree_tests/
    ```

*
    ```bash
    # This keeps the conftest.py at the root, which might not be wanted
    iree_tests/
      common/
      onnx/
      pytorch/
      special_models/
    ```

*
    ```bash
    # This excludes special_tests from conftest.py. Need a name for non-special tests
    iree_tests/
      common/
      tests/
        onnx/
        pytorch/
      special_tests/
        sd3/
    ```

Naming is hard :P

Edit: Could keep `ireers/` instead of `common/` (or nested under `common/`, to keep it installable.

iree_special_models/sdxl/prompt-encoder/test_clip.py
```diff
@@ -0,0 +1,111 @@
...
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import pytest
+from ireers import *
+import os
+
+# iree_tests/pytorch/complex_models/sdxl/prompt-encoder
```
ScottTodd, line 11: What's this comment? Should it be updated to reflect the current path?

iree_special_models/sd3/mmdit/test_mmdit.py
```diff
@@ -0,0 +1,120 @@
...
+
+CPU_COMPILE_FLAGS = [
+    "--iree-hal-target-backends=llvm-cpu",
+    "--iree-llvmcpu-target-cpu-features=host",
+    "--iree-llvmcpu-fail-on-out-of-bounds-stack-allocation=false",
+    "--iree-llvmcpu-distribution-size=32",
+    "--iree-opt-const-eval=false",
+    "--iree-llvmcpu-enable-ukernels=all",
+    "--iree-global-opt-enable-quantized-matmul-reassociation"
+]
```
ScottTodd, line 28: (Sorry, thoughts bouncing around here :P)

I guess for testing in IREE we'd consider forking this entire file so these options could be overridden and new configurations could be added (e.g. Vulkan, CUDA), but at that point we'd be just sharing
* the 90 lines from ireers/fixtures.py
* The method for downloading files

So I'm wondering if it would make more sense to fold changes into https://github.com/iree-org/iree/tree/main/experimental/regression_suite and get that out of `experimental/`. The file downloading _is_ unique to nod-ai/SHARK-TestSuite though... IREE upstream would prefer to directly use some neutral hosted location instead of the shark Azure buckets.

... (insert thinking noises here) ...

How about this: we keep this PR mostly as-is and treat it as a "community CI build" in IREE. AMD is supplying the runners, data, and test files, and all of that configuration exists here in this repo. As the move into the Linux Foundation progresses and this test suite gets some mileage we can consider moving it to iree-org/, using some neutral cloud hosting, and including models (and compile flags, etc.) that other contributors care about.

It _would_ still be useful to be able to change e.g. `ROCM_COMPILE_FLAGS` from IREE though - especially if a flag is renamed there... we wouldn't want the test to fail with no way to fix it that doesn't cross repos.

Thoughts?

.github/workflows/test_iree.yml
```diff
@@ -162,57 +164,37 @@ jobs:
...
-      - name: "Running real weights SDXL prompt encoder tests"
-        id: prompt_encoder
+      
+      - name: "Running SDXL special model tests"
+        id: special_models_sdxl
         if: ${{ !cancelled() }}
         run: |
           source ${VENV_DIR}/bin/activate
-          pytest iree_tests/pytorch/models/sdxl-prompt-encoder-tank \
+          pytest iree_special_models/sdxl \
```
ScottTodd, line 173: Assuming that the new directories don't end up nested under `iree_tests`, they should be added to the 'paths' that trigger this workflow https://github.com/nod-ai/SHARK-TestSuite/blob/4751fab06fdc08818ee30530e0cdfc38fc6ce4b7/.github/workflows/test_iree.yml#L7-L16

.github/workflows/test_iree.yml
```diff
@@ -99,25 +100,19 @@ jobs:
       matrix:
         include:
           - name: cpu_llvm_task
-            runs-on: nodai-amdgpu-w7900-x86-64
+            runs-on: nodai-amdgpu-mi250-x86-64
```
ScottTodd, line 103: Looks like I'll need this change too. Might want to separate changes like these into their own PRs.

iree_special_models/sd3/mmdit/test_mmdit.py
saienduri: Yeah your thoughts are basically what I was thinking too. It would be good to have the flags changeable in IREE. My main thing was I wanted to eliminate all those json files with this implementation, so I just went with this. The one way I thought was maybe having a special_models common json file. We would add all the compile flags for all the models. But, then I thought this would be too messy as we add models (can still do that, just not sure how sustainable/neccessary it is). If we split it up based on model, it would just end up like the original situation with a bunch of json files.  I think the main thing is I couldn't think of a sustainable way where we would not have to add a bunch of files. Open for any option that doesn't require a bunch of files though.

iree_special_models/ireers/fixtures.py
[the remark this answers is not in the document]
saienduri: Yeah, still thought the name was fitting though

iree_special_models/ireers/fixtures.py
```diff
@@ -0,0 +1,90 @@
...
+    compile_args = [
+        "iree-compile",
+        mlir_path
+    ]
+    compile_args += args
+    compile_args += ["-o", output_file]
+    cmd = subprocess.list2cmdline(compile_args)
+    return cmd
+
+def iree_compile(mlir_path: str, output_file: str, args: [str], cwd):
```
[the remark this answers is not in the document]
saienduri: We run the subprocess command with the env parameter being in cwd, so these can just be file names or the path of the files (went with string to make sense of all cases).

iree_tests/requirements.txt
```diff
@@ -6,6 +6,7 @@ pytest-timeout
 pytest-xdist
 pyjson5
 pytest-check
+pytest-depends
```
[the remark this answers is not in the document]
saienduri: Created another requirements for special models. Weird with only one requirement now lol, but I guess it will grow in future

iree_special_models/sd3/mmdit/test_mmdit.py
ScottTodd: Yeah this is definitely an improvement. Forking the .py files (to files like https://github.com/iree-org/iree/blob/main/experimental/regression_suite/tests/pregenerated/test_llama2.py) may be reasonable. What I'm watching out for is:

* IREE contributors without access to these Azure storage accounts not getting stuck if the files need to be regenerated (so need a way to either contribute updates or at least disable tests temporarily)
* IREE contributors need a way to change flags and configurations (add new ones for testing, remove flags after changing them, etc.)
* Complexity for:
  * Adding new tests
  * Seeing which tests are defined
  * Seeing which flags are used
  * Checking which tests are passing/failing
  * Adding benchmarks

We're talking about ~300 lines of code per test model (across multiple configurations), so ~5000 lines of code total across ~15 test models. That's not _too_ bad for maintenance.

> The one way I thought was maybe having a special_models common json file. We would add all the compile flags for all the models on cpu and rocm.

Actually for this, we can have common python files, so that might not be so bad.

```python
DEFAULT_ROCM_FLAGS = [ "foo", "bar" ]
WINOGRAD_ROCM_FLAGS = ...
SDXL_ROCM_FLAGS = DEFAULT_ROCM_FLAGS + ...

ALL_BACKENDS = ...

def register_tests_for_model(model=..., excluded_backends=[...])
```

if we add that sort of configuration then it should really live in the IREE repo and not here. What we _could_ store here would be model importing tools, sanity check tests, git LFS files, or URLs to huggingface / Azure / GCS / etc.

saienduri: Cool, this sounds good to me. Can we have the test files live here and in iree? I think there is a value to be able to run all tests out of here (e2e, iree normal models, iree special models). IREE can use the TestSuite and have its own test files so easy to make changes there, but I would like the TestSuite to also be something that you can clone and be able to run tests on everything SHARK related.

ScottTodd: Yeah, I guess let's be pragmatic. We can keep a copy here and in IREE and just sync them manually as needed.

If we have only one place exhaustively test configurations, I'd prefer for that to be IREE (to catch regressions and measure performance using code on PRs and landed commits).

iree_special_models/sd3/vae-decode/test_cases.json
```diff
@@ -4,8 +4,8 @@
       {
         "name": "real_weights",
         "remote_files": [
-          "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd3-vae/inference_input.0.bin",
-          "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd3-vae/inference_output.0.bin",
+          "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd3-vae-updated/inference_input.0.bin",
+          "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd3-vae-updated/inference_output.0.bin",
```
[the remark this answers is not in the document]
saienduri: Haha yeah, good catch, I was just testing if it works. When finally landing, I will just overwrite the original locations :)

iree_special_models/sd3/mmdit/test_mmdit.py
saienduri: Ok yeah makes sense

.github/workflows/test_iree.yml
```diff
@@ -63,6 +65,8 @@ jobs:
         run: |
           source ${VENV_DIR}/bin/activate
           python3 -m pip install -r iree_tests/requirements.txt
+          python3 -m pip install -r iree_special_models/requirements.txt
```
ScottTodd, line 68: Are these extra requirements needed in this job? This is just running onnx tests here

.github/workflows/test_iree.yml
```diff
@@ -140,13 +139,16 @@ jobs:
...
 
       # Download remote files.
       - name: "Downloading remote files for real weight model tests"
         run: |
           source ${VENV_DIR}/bin/activate
-          python3 iree_tests/download_remote_files.py --root-dir pytorch/models
-          python3 iree_tests/download_remote_files.py --root-dir sharktank
+          python3 iree_tests/download_remote_files.py --root-dir iree_tests/pytorch/models
+          python3 iree_tests/download_remote_files.py --root-dir iree_special_models
+          python3 iree_tests/download_remote_files.py --root-dir iree_tests/sharktank
```
ScottTodd, line 151: nit: sort (could put special models at the top for alphabetical, or put lower)

```suggestion
          python3 iree_tests/download_remote_files.py --root-dir iree_special_models
          python3 iree_tests/download_remote_files.py --root-dir iree_tests/pytorch/models
          python3 iree_tests/download_remote_files.py --root-dir iree_tests/sharktank
```

We may also want to make the cache scanning faster as we add more files

iree_special_models/sd3/mmdit/test_mmdit.py
```diff
@@ -0,0 +1,120 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import pytest
+from ireers import *
+import os
+
+repo_root = os.getenv("TEST_SUITE_REPO_ROOT")
+current_dir = repo_root + "/iree_special_models/sd3/mmdit"
```
ScottTodd, line 12: This is brittle (if the file is moved, this line will also need to be updated). How about either `os.path.dirname(__file__)` or `Path(__file__).parent`?

.github/workflows/test_iree.yml
```diff
@@ -164,64 +166,44 @@ jobs:
...
             --log-cli-level=info \
             --timeout=1200 \
-            --durations=0 \
-            --config-files=${SDXL_PROMPT_ENCODER_CONFIG_FILE_PATH}
-
-      - name: "Running real weights SDXL scheduled unet tests"
-        id: unet
+            --durations=0
+      
+      - name: "Running SD3 special model tests"
```
ScottTodd, line 183: This CPU job took 26 minutes: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/9748891096/job/26904892398?pr=271

seems like a lot...

* Another run on of my PRs: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/9719565868/job/26829591994 
* A run in IREE: https://github.com/iree-org/iree/actions/runs/9748314047/job/26904123071

where is the time coming from? We should aim for 20-30 minutes total on presubmit (including 10 minutes building + time queued for runners). Longer durations should be split into multiple jobs or moved to a `schedule` instead of on `pull_request`

ScottTodd (approved):
Just a few more questions then looks good to land

.github/workflows/test_iree.yml
saienduri: This is the difference between w7900 and mi250. w7900 is super fast. I will try to get that machine back in here but that one is pretty unstable

iree_special_models/sd3/mmdit/test_mmdit.py
saienduri: Yeah that was how it was before, but to be able to have this implemented downstream we will have to go with this for now I think. Because, these python files will be in different places than where we are cloning the SHARK-TestSuite repo right?

.github/workflows/test_iree.yml
saienduri: Not needed, removed

iree_special_models/sd3/mmdit/test_mmdit.py
ScottTodd: Consider how much code is forked vs used directly then decide on a setup from there.

This is only used here to get the path to `attention_and_matmul_spec.mlir`. If the entire test file is forked (so any flags and configurations can be changed), then there's no need to try making this something that can be substituted.

saienduri: This is also used for all the calls to iree_compile and iree_run_module where we specify the directory to run the command from

PR #273, opened 2024-06-28
Cleanup/refactoring in iree_tests.

ScottTodd:
* Drop "config_" prefix from iree_tests config files.
* Qualify test names relative to iree_tests/.
* Switch CPU model tests to run on mi250.
* Use `less_equal` instead of `less` for benchmark dispatch binary size comparison.


saienduri (approved):
LGTM

iree_tests/conftest.py
```diff
@@ -246,26 +248,29 @@ def collect(self):
         #     ...
 
         test_directory = self.path.parent
+        relative_test_directory = test_directory.relative_to(IREE_TESTS_ROOT).as_posix()
```
ScottTodd, line 251: Ah, need to update https://github.com/nod-ai/SHARK-TestSuite/blob/main/iree_tests/update_config_xfails.py also

ScottTodd: Fix: https://github.com/nod-ai/SHARK-TestSuite/pull/280

PR #274, opened 2024-06-28
Teach download_remote_files.py to use huggingface_hub.

ScottTodd:


saienduri (approved):
LGTM

ScottTodd:
Oops, description from another PR:


> As part of this, I added support for downloading models from Hugging Face (in addition to Azure Storage). The `huggingface_hub` package has its own caching strategy, which I adapted to work with our custom caching for our GitHub Actions runners.
> * Hugging Face limits access to certain files based on authentication (e.g. the `HF_TOKEN` env var), which we already handle in the `e2eshark/` subproject. If we add tests for gated models then we can set that up here somehow too.
> * Generic file downloading is still unimplemented, and as a reminder we can add other hosts like Google Cloud Storage, Amazon S3, etc. as needed.

PR #277, opened 2024-07-02
Import punet tests from sharktank to iree_tests.

ScottTodd:
Related to https://github.com/nod-ai/sharktank/issues/76

* Model source: https://github.com/nod-ai/sharktank/tree/main/sharktank/sharktank/models/punet
* Python test: https://github.com/nod-ai/sharktank/blob/main/sharktank/integration/models/punet/integration_test.py
* Python run script: https://github.com/nod-ai/sharktank/blob/main/sharktank/sharktank/models/punet/tools/run_punet.py
* Weights are labeled in the sharktank sources. The primarily come from https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0 and https://huggingface.co/amd-shark/sdxl-quant-models

Only testing compilation and basic execution for now. Result comparisons will follow once they are enabled upstream. TBD if the quantized outputs will be different enough across backends to need extra infrastructure configuration there.

PR #278, opened 2024-07-03
Addition of 4 pytorch onnx-export models.

saienduri:
This commit adds testing for 4 models that we have noticed to be unstable and want to keep passing through CI:

mit-b0
mobilebert-uncased
t5-base
t5-large

These models are run through the onnx mode and pytorch framework where the pytorch model is exported to onnx, imported to torch IR using the onnx importer, and then fed to IREE to lower to linalg and run the module.

We currently don't support the use of external parameters through the onnx flow thus the lack of parameter files and splat testing

Also added the --iree-input-demote-i64-to-i32 flag to the cpu models config file so that we are in line with what we run in e2eshark for any cpu code generation


iree_tests/configs/models_cpu_llvm_task.json
```diff
@@ -2,20 +2,19 @@
     "config_name": "cpu_llvm_task",
     "iree_compile_flags" : [
       "--iree-hal-target-backends=llvm-cpu",
-      "--iree-llvmcpu-target-cpu-features=host"
+      "--iree-llvmcpu-target-cpu-features=host",
+      "--iree-input-demote-i64-to-i32"
     ],
```
ScottTodd, line 7: > Also added the --iree-input-demote-i64-to-i32 flag to the cpu models config file so that we are in line with what we run in e2eshark for any cpu code generation

I'm not sure I agree with this. I'm worried about the number of flags we have developers using, and I'd like to have more visibility into which models require which flags and why. One way we can do that is by having users of the test suite test with default flags.

In this case, the `--iree-input-` flags are sort of special in that they are generic across backends so I do see a case for using some of them... maybe in a `_compatibility_flags` test suite config, then we could diff the XFAILs between that and the `_default_flags` config.

iree_tests/pytorch/models/onnx-export/mit-b0/model.mlirbc
ScottTodd, line 1: > These models are run through the onnx mode and pytorch framework where the pytorch model is exported to onnx, imported to torch IR using the onnx importer, and then fed to IREE to lower to linalg and run the module.

That seems rather roundabout... I guess ONNX there is used as a serialization format?

IMO if we're adding tests for these models we should add _both_ the direct and onnx-export versions.

iree_tests/configs/models_cpu_llvm_task.json
```diff
@@ -2,20 +2,19 @@
...
     "expected_compile_failures": [
-      "pytorch/models/opt-125M", // TODO(#17344): need to regenerate .mlirbc
-      "pytorch/models/resnet50", // TODO(#17344): need to regenerate .mlirbc
+      "pytorch/models/direct/opt-125M", // TODO(#17344): need to regenerate .mlirbc
+      "pytorch/models/direct/resnet50", // TODO(#17344): need to regenerate .mlirbc
     ],
-    "expected_run_failures": []
+    "expected_run_failures": [
+      "pytorch/models/onnx-export/mobilebert-uncased",
+    ]
```
ScottTodd, line 19: > This commit adds testing for 4 models that we have noticed to be unstable and want to keep passing through CI

Can we add models that we think are _stable_ too? :) I'd like to prefetch all models from e2eshark and get them running here if possible, rather than cherrypick a few _after_ we notice issues.

iree_tests/configs/models_gpu_rocm_gfx90a.json
```diff
@@ -2,25 +2,23 @@
...
-    ],
+    "skip_compile_tests": [],
     "skip_run_tests": [],
     "expected_compile_failures": [
-      "pytorch/models/opt-125M", // TODO(#17344): need to regenerate .mlirbc
-      "pytorch/models/resnet50",
+      "pytorch/models/direct/opt-125M", // TODO(#17344): need to regenerate .mlirbc
+      "pytorch/models/direct/resnet50",
       // error: 'builtin.module' op failed to run transform dialect passes
       // (might need to drop the iree-codegen-transform-dialect-library flag)
```
ScottTodd, line 16: Can drop this comment since the `attention_and_matmul_spec` is removed.
I also saw `error: a handle passed as operand #0 and consumed by this operation points to a payload entity more than once` and updated the comment in my draft PR: https://github.com/nod-ai/SHARK-TestSuite/pull/277

How about we just leave the reason comments off? Also on my PR there, if we run with `-rA` (instead of `-rpfE`, see https://docs.pytest.org/en/stable/how-to/output.html), we'll get full error output for even XFAIL'd tests in CI logs, so having an possibly outdated comment in the source isn't as useful.

iree_tests/pytorch/models/onnx-export/t5-base/test_cases.json
```diff
@@ -0,0 +1,65 @@
...
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-base/inference_output.43.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-base/inference_output.44.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-base/inference_output.45.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-base/inference_output.46.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-base/inference_output.47.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-base/inference_output.48.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-base/inference_output.49.bin",
```
ScottTodd, line 60: o_o that's a lot of outputs. Fine for now, but let's file an issue to use archive files (.zip or .tar.gz). If we use something like Hugging Face repositories, we could also lean on their download API to fetch an entire repo

iree_tests/pytorch/models/onnx-export/t5-large/test_cases.json
```diff
@@ -0,0 +1,113 @@
...
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-large/inference_output.91.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-large/inference_output.92.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-large/inference_output.93.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-large/inference_output.94.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-large/inference_output.95.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-large/inference_output.96.bin",
+        "https://sharkpublic.blob.core.windows.net/sharkpublic/shark-test-suite/iree-tests/pytorch/models/onnx-export/t5-large/inference_output.97.bin",
```
ScottTodd, line 108: Oh, that's even more outputs :P

ScottTodd (commented):
Thanks! Generally looks great. A few ideas for improvement, but could proceed without all addressed.

ScottTodd:
This PR looks stale, closing. Let's focus IREE tests in https://github.com/iree-org/iree-test-suites now.

PR #282, opened 2024-07-08
Refactor pytest collection to anchor on test case files, not .mlir files.

ScottTodd:
This allows us to detect test cases that store their `.mlir` or `.mlirbc` file remotely, since the `test_data_flags.txt` or `test_cases.json` file will always exist locally. This also changes test case names in summaries from e.g. 
```
PASSED onnx/node/generated/test_sub_uint8/model.mlir::cpu_llvm_sync_test
PASSED onnx/node/generated/test_sub_example/model.mlir::cpu_llvm_sync_test
PASSED onnx/node/generated/test_sub_bcast/model.mlir::cpu_llvm_sync_test

XFAIL pytorch/models/opt-125M/opt-125M.mlirbc::cpu_llvm_task_splats - Expected compilation to fail (included in 'expected_compile_failures')
```
to e.g.
```
PASSED onnx/node/generated/test_sub_example/test_data_flags.txt::model.mlir::cpu_llvm_sync
PASSED onnx/node/generated/test_sub_uint8/test_data_flags.txt::model.mlir::cpu_llvm_sync
PASSED onnx/node/generated/test_sub_bcast/test_data_flags.txt::model.mlir::cpu_llvm_sync

XFAIL pytorch/models/opt-125M/test_cases.json::opt-125M.mlirbc::cpu_llvm_task::splats - Expected compilation to fail (included in 'expected_compile_failures'
```
(anchored on the .txt or .json so added back the .mlir, also dropped the "test")

This is a bit hacky either way. I took a look at https://docs.pytest.org/en/stable/example/customdirectory.html as an alternative to https://docs.pytest.org/en/stable/example/nonpython.html and that could help.

### DIFF
--- a/iree_tests/conftest.py
+++ b/iree_tests/conftest.py
@@ -16,6 +16,7 @@ import subprocess
 from ireers import *
 
 IREE_TESTS_ROOT = Path(__file__).parent
+TEST_DATA_FLAGFILE_NAME = "test_data_flags.txt"
 
 # --------------------------------------------------------------------------- #
 # pytest hooks
@@ -112,10 +113,14 @@ def pytest_sessionstart(session):
 
 
 def pytest_collect_file(parent, file_path):
-    if not file_path.name.endswith("_spec.mlir") and (
-        file_path.name.endswith(".mlir") or file_path.name.endswith(".mlirbc")
-    ):
-        return MlirFile.from_parent(parent, path=file_path)
+    if file_path.name == TEST_DATA_FLAGFILE_NAME:
+        return MlirCompileRunTest.from_parent(parent, path=file_path)
+
+    if file_path.suffix == ".json":
+        with open(file_path) as f:
+            test_cases_json = pyjson5.load(f)
+            if test_cases_json.get("file_format", "") == "test_cases_v0":
+                return MlirCompileRunTest.from_parent(parent, path=file_path)
 
 
 # --------------------------------------------------------------------------- #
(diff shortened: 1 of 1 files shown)


iree_tests/conftest.py
```diff
@@ -206,45 +212,54 @@ def discover_test_cases(self):
...
-                MlirFile.TestCase(
-                    name="test",
-                    runtime_flagfile=test_data_flagfile_name,
+                MlirCompileRunTest.TestCase(
+                    name="",
+                    mlir_file=mlir_file,
+                    runtime_flagfile=TEST_DATA_FLAGFILE_NAME,
                     enabled=have_lfs_files,
                 )
             )
```
ScottTodd, line 233: Could remove the need for this branch by generating boilerplate `test_cases.json` files in the ONNX test suite:

```json
{
  "file_format": "test_cases_v0",
  "test_cases": [
    {
      "name": "",
      "runtime_flagfile": "test_data_flags.txt",
      "remote_files": []
    }
  ]
}
```

Trying to keep this conftest file _somewhat_ simple and a bit concerned this is sliding the wrong way 🤔

saienduri: Yeah, would probably be best to have test_cases.json files for all the onnx tests too. That way all our tests can follow a unified format and the conftest won't get confusing. We should probably be anchoring on a test configuration file rather than the runtime flag file for onnx tests

ScottTodd: Okay, I can make that change too. How does a follow-up PR sound? That will require touching lots of files.

saienduri (approved):
Follow up is fine

iree_tests/conftest.py
ScottTodd: Might also be able to replace `test_cases.json` with `tests.py` and then further simplify (or remove) the `conftest.py` file... 🤔 Mainly trying to keep the large test suites autogenerated, concise, and directly compatible with `iree-compile` -> `iree-run-module`.

PR #284, opened 2024-07-09
Adjust schedule for test_e2eshark nightly run.

ScottTodd:
This job is hogging the CI runner during PST working hours: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/9856657975/job/27214130493, so start it earlier in the morning.

The workflow history also shows the runs taking 5-6 hours but only running for 3 hours? https://github.com/nod-ai/SHARK-TestSuite/actions/workflows/test_e2eshark.yml. Is the machine offline or busy during that time? Moving the schedule might not help in that case.

### DIFF
--- a/.github/workflows/test_e2eshark.yml
+++ b/.github/workflows/test_e2eshark.yml
@@ -8,8 +8,8 @@ name: E2ESHARK Test Suite
 on:
   workflow_dispatch:
   schedule:
-  # Runs at 12:00 PM UTC, which is 5:00 AM PST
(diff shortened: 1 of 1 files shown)


saienduri:
Something fishy here, it should only be taking 3 hours. Not sure why it is hogging for an extra 3 hours. Will stop and start up the service again to see if that makes a difference tonight

ScottTodd:
I either haven't seen similar issues recently or they've been minor enough to ignore. Closing this.

PR #307, opened 2024-07-29
[alt] make `--cachedir` optionally specifiable from an env. variable

zjgarvey:
Now, instead of needing to do:
```
python run.py --cachedir=/path/to/cachedir <otheroptions>
```
You can set 
```
export CACHE_DIR="/path/to/cachedir"
```
Then,
```
python run.py <otheroptions>
```

This PR also adds a convenience class for re-running the same test with different inputs or post-processing steps.


alt_e2eshark/run.py
```diff
@@ -66,8 +67,17 @@ def get_tests(groups, test_filter):
...
-    # TODO: add better logging setup
+
+    # set-up cache directory. Tries to get from env. var.
+    CACHE_DIR = os.getenv('CACHE_DIR')
+    cache_dir = args.cachedir
+    if not CACHE_DIR and not cache_dir:
+        raise RuntimeError("No CACHE_DIR environment variable set, and no --cachedir arg provided.")
+    if cache_dir:
+        # if a --cachedir arg is provided, use it for the tests and update the env variable.
+        os.environ['CACHE_DIR'] = cache_dir
```
renxida, line 78: I don't like this because typically linux commands with arguments like this overrides the env var for a single run and does not update the env variable.

To allow the user to set a default for all future runs, I would much rather read a bunch of variables called something like SHARKTEST_XXX and have users put them into a .env file.

renxida (changes requested):
Love env variables, don't love setting them from a cmdline arg.

alt_e2eshark/run.py
renxida: making --cachedir set the environment variable might cause surprises & make for weird persistence issues. I think we should let users manage their own environment variables and include info on the environment variables in the --help output

renxida (approved):

PR #314, opened 2024-08-02
Changes for CI Test File Usage

saienduri:
Changes so that the CI uses a test file and moves to using the mi250 Alpha 2 node cluster, so some caching changes for that as well


.github/workflows/test_e2eshark.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     runs-on: nodai-amdgpu-mi250-x86-64
     env:
       E2E_VENV_DIR: ${{ github.workspace }}/test-suite_venv
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
+      AZ_PRIVATE_CONNECTION: ${{ secrets.AZ_PRIVATE_CONNECTION }}
```
ScottTodd, line 20: Please qualify this with which storage account this private connection string is used for, e.g. `ONNXPRIVATESTORAGE_AZ_PRIVATE_CONNECTION`. I wasn't sure what needed updating in this repo after I rotated keys last time.

.github/workflows/test_e2eshark.yml
```diff
@@ -57,14 +59,18 @@ jobs:
 
       - name: Run Onnx Mode
         run: |
+          echo $AZ_PRIVATE_CONNECTION
+          echo $HF_TOKEN
+          echo $E2E_VENV_DIR
```
ScottTodd, line 64: Remove these lines. https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#accessing-your-secrets

> Warning: If a secret was used in the job, GitHub automatically redacts secrets printed to the log. You should avoid printing secrets to the log intentionally.

e2eshark/ci-list.txt
ScottTodd, line 1: This should be documented - maybe in e2eshark/README.md?

* We would run all tests on CI if we had the resources / capacity for it (with tests being sharded across multiple machines)
* Until we do have those resources, only run a limited subset

.github/workflows/test_e2eshark.yml
saienduri: Thanks, forgot that was even there :)

e2eshark/ci-list.txt
saienduri: Yeah, there will be changes coming to this in the next week. Going to focus on getting all the models running, so will update the README after that

saienduri:
Merging so the nightly runs keep running

PR #315, opened 2024-08-02
Change types for some tests to correct types

rsuderman:


iree_tests/onnx/node/generated/test_cast_BFLOAT16_to_FLOAT/test_data_flags.txt
```diff
@@ -1,2 +1,2 @@
---input=3x4xui16=@input_0.bin
+--input=3x4xbf16=@input_0.bin
```
ScottTodd, line 1: These files are generated. The type conversion from onnx happens here: https://github.com/nod-ai/SHARK-TestSuite/blob/52468d636519c6409f5f28623357f218b92a369e/iree_tests/onnx/import_tests_utils.py#L10-L24 https://github.com/nod-ai/SHARK-TestSuite/blob/52468d636519c6409f5f28623357f218b92a369e/iree_tests/onnx/import_tests.py#L142-L147

is that code not handling a case here? I'd expect the `WARNING: unsupported data type in get_shape_string` to be hit 🤔

ScottTodd: While we _can_ run the generator once and then manually update tests like this, I don't want for future runs of the generator to need to manually make these same changes, so I'd prefer for the generator to be taught to make these changes itself.

If we're losing information across the onnx -> numpy -> iree translation, we could add some special cases in the generator script (similar to how I suggested adding per-test tolerances on https://github.com/nod-ai/SHARK-TestSuite/pull/311#discussion_r1700604348)

rsuderman: I would need to dig into the proto but my guess is that they just identify themselves at uint16 with an internal bitcast. `bf16` is a bit weird because from a data standpoint its just a truncated `f32` and is basically treated that way. This results in buffers often just being passed around at `uint16`

ScottTodd:
Oh, can you sync past https://github.com/nod-ai/SHARK-TestSuite/commit/53bedfbd066b5b7b181526b956daded98539b7f7 ? The important detail there is the `skip_compile_tests` additions (to avoid timeouts).

PR #316, opened 2024-08-05
[onnx] Fix test generator for better type information

rsuderman:
Test generator does not handle all type imports. Fixed the test generator so that we actually convert the types correctly.

### DIFF
--- a/iree_tests/onnx/import_tests.py
+++ b/iree_tests/onnx/import_tests.py
@@ -54,6 +54,64 @@ def convert_io_proto(proto_filename, type_proto):
             return None
 
 
+def convert_proto_etype(etype):
+    if etype == onnx.TensorProto.FLOAT:
+        return "f32"
+    if etype == onnx.TensorProto.UINT8:
+        return "ui8"
+    if etype == onnx.TensorProto.INT8:
+        return "si8"
+    if etype == onnx.TensorProto.UINT16:
+        return "ui16"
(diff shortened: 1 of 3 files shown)


iree_tests/onnx/import_tests.py
```diff
@@ -140,32 +200,26 @@ def import_onnx_files(test_dir_path, imported_dir_path):
     for i in range(len(test_inputs)):
         test_input = test_inputs[i]
         t = convert_io_proto(test_input, model.graph.input[i].type)
+        ty = get_io_proto_type(model.graph.input[i].type)
         if t is None:
             return False
         input_path = (imported_dir_path / test_input.stem).with_suffix(".npy")
         np.save(input_path, t)  # Only for ref, actual comparison with .bin
-        ss = get_shape_string(t)
```
ScottTodd, line 147: If this is no longer used, please remove it from https://github.com/nod-ai/SHARK-TestSuite/blob/main/iree_tests/onnx/import_tests_utils.py 

You could put `convert_proto_etype` and `get_io_proto_type` in that file too, so this file is smaller... or merge the two files together. I don't have a strong preference.

ScottTodd (approved):

iree_tests/onnx/import_tests.py
rsuderman: Done.

PR #321, opened 2024-08-09
Fix Path instruction in README.md

vivekkhandelwal1:


iree_tests/README.md
```diff
@@ -88,7 +88,7 @@ To use local versions of `iree-compile` and `iree-run-module`, put them on your
 `$PATH` ahead of your `.venv/Scripts` directory:
 
 ```bash
-$ export PATH=path/to/iree-build;$PATH
+$ export PATH=path/to/iree-build:$PATH
 ```
```
ScottTodd, line 92: Ah, I wrote this for Windows, where the path separator is `;`. It is indeed `:` on Linux. Fine to use Linux syntax here, as that matches the other docs (venv setup is a bit different on Windows too, for example). In more expressive markdown / doc formats I'll use content tabs like those here: https://iree.dev/reference/bindings/python/#prerequisites

Linux selected: 
image

Windows selected:
image

ScottTodd (approved):

PR #313, opened 2024-08-02
[ONNX] Add HF CNN fp32 model tests

jinchen62:
| items        |    tests |   model-run |   onnx-import |   torch-mlir |   iree-compile |   inference |
|:-------------|---------:|------------:|--------------:|-------------:|---------------:|------------:|
| total-count  | 530 |     500     |       500     |            129 |        73     |     16     |

full list (530): https://gist.github.com/jinchen62/cdf54ef8ed725fcce9d6fa18ecbfa058
failed on model-run (30): https://gist.github.com/jinchen62/1b23c9c227b772c7e7641fad39aa6ea9
failed on torch-mlir/iree-compile (411): https://gist.github.com/jinchen62/0ad6931179badcba0860f2da01b3cc21
failed on inference (73): https://gist.github.com/jinchen62/c870373258c5a1c71c975f81258463de

commands:
`python ./run.py  --tolerance 0.001 0.001 -c path/to/torch-mlir/build -i path/to/iree-build --cachedir path/to/cache -f onnx -g models --mode onnx --report -j 12 --testsfile model_list.txt`

PR #322, opened 2024-08-12
[ONNX] Remove darknet53 model test

jinchen62:
DarkNet53 exists, the only difference is input shape.

Address https://github.com/nod-ai/SHARK-TestSuite/pull/313#discussion_r1712041920


ScottTodd (approved):
Thanks!

PR #329, opened 2024-08-20
Update opset version to 21 when importing onnx models

jinchen62:


zjgarvey (approved):
Seem reasonable to me, but I'm sure there will be some models that fail to update to version 21 properly.

zjgarvey:
Also, please change the title to something reflective of the change. E.g., "update opset version to 21 when importing"

PR #330, opened 2024-08-22
[ONNX][EP] Add provider_options parameter in IREE-EP backend

Shukla-Gaurav:


zjgarvey:
Nice, I imagine you can use the existing cl flags `--device` and `--backend` and `--iree-compile-args` to set the provider options in the init for the config class rather than passing the provider options directly.

zjgarvey:
lmk when this is ready for review. If it's a [DO NOT MERGE] you might as well make it a draft PR.

zjgarvey (approved):
Yeah, this looks great. I assume you've used it to test on CPU/GPU and with various flags, right? If it works as expected, lgtm.

Shukla-Gaurav:
> Yeah, this looks great. I assume you've used it to test on CPU/GPU and with various flags, right? If it works as expected, lgtm.

Yes, it works fine. I have pushed few more changes, like sorting the test list(it helps when the execution breaks due to segfault to keep track of models that passed). Could you please take a look?

alt_e2eshark/e2e_testing/backends.py
```diff
@@ -145,26 +145,49 @@ def func(x: TestTensors) -> str:
...
+                "--iree-input-demote-i64-to-i32",
+                # "--iree-llvmcpu-fail-on-large-vector=0",
+                # "--iree-llvmcpu-stack-allocation-limit=300000",
+            ]
+        self.providers = ["IreeExecutionProvider"]
+        # set provider options.
+        provider_options_dict = dict()
+        provider_options_dict["hal_target_device"] = self.hal_target_device
+        provider_options_dict["device"] = self.device
+        provider_options_dict["compile_time_flags"] = "+".join(self.extra_args)
```
zjgarvey, line 174: Are the compile time flags separated by `+`? This seems kinda odd.

zjgarvey (approved):
This looks good to me.

alt_e2eshark/e2e_testing/backends.py
Shukla-Gaurav: Yes. There can be special characters in the compile time flags. For now, added `+` as the separator, will come with better combination of characters as delimiter. Thanks!

PR #337, opened 2024-09-12
Use a separate py file to register nlp models and provide an appropriate dim_param_dict with some sane defaults

renxida:


alt_e2eshark/e2e_testing/onnx_utils.py
```diff
@@ -39,6 +39,10 @@ def generate_input_from_node(node: onnxruntime.capi.onnxruntime_pybind11_state.N
                 int_dims.append(dim_param_dict[dim])
                 continue
         if not isinstance(dim, int):
+            if isinstance(dim, str):
+                print(f"generate_input_from_node: WARNING: filling in placeholder value 16 for input named `{dim}`. Consider setting custom inputs for this test.")
```
zjgarvey, line 43: I'd say "consider setting a self.dim_param_dict for this test".

alt_e2eshark/e2e_testing/onnx_utils.py
```diff
@@ -39,6 +39,10 @@ def generate_input_from_node(node: onnxruntime.capi.onnxruntime_pybind11_state.N
...
+                int_dims.append(16)
+                continue
             raise TypeError(
                 f"input node '{node.name}' has dims={node.shape}. Node dim '{dim}' has invalid type: {type(dim)}\nexpected type: int.\nIf your model has dim_params, consider fixing them or setting custom inputs for this test."
```
zjgarvey, line 47: I'd change this error message since it is no longer relevant. Something like 
```python
f"input node '{node.name}' has dim `{dim}` of unsupported type '{type(dim)}'."
```

zjgarvey (changes requested):
I'd change some of the messaging. 

I'd also recommend providing appropriate `dim_param_dicts` for each test with dynamic dims. See some of the migraphx tests registration, for instance.

renxida:
Tested and was able to get to some numerical failures!!!!

zjgarvey (approved):
LGTM for now. We can iterate on this in the future, but we need to get these in the CI today.

zjgarvey:
Thanks!

PR #341, opened 2024-09-17
Update README.md

kumardeepakamd:
Copy paste of extra / at end errors out, we should have instructions that can be copy pasted for ease of use


saienduri (approved):
Thanks

PR #342, opened 2024-09-17
[alt] Add a CL-based mode

zjgarvey:
Adds `--mode=cl-onnx-iree` flag option which mirrors `--mode=onnx-iree` (default), but runs most stages via command line scripts rather than python bindings. This seems to work better for a few reasons:

1. Easier command-line reproducers for failures. (The script which fails is printed in the corresponding log file)
2. Memory management per stage is more contained and doesn't need to rely on python's garbage collector.

### DIFF
--- a/alt_e2eshark/e2e_testing/backends.py
+++ b/alt_e2eshark/e2e_testing/backends.py
@@ -115,7 +115,7 @@ class CLIREEBackend(BackendBase):
         for arg in self.extra_args:
             arg_string += arg
             arg_string += " "
-        command_error_dump = os.path.join(save_to, "compilation.detail.log")
+        command_error_dump = os.path.join(save_to, "detail/compilation.detail.log")
         script = f"iree-compile {module_path} {arg_string}-o {vmfb_path} 2> {command_error_dump}"
         os.system(script)
         if not os.path.exists(vmfb_path):
--- a/alt_e2eshark/e2e_testing/backends.py
+++ b/alt_e2eshark/e2e_testing/backends.py
@@ -74,7 +74,7 @@ class SimpleIREEBackend(BackendBase):
         )
         # log the vmfb
         if save_to:
-            with open(save_to + "compiled_model.vmfb", "wb") as f:
+            with open(os.path.join(save_to, "compiled_model.vmfb"), "wb") as f:
                 f.write(b)
         return b
 
@@ -110,14 +110,14 @@ class CLIREEBackend(BackendBase):
                     self.extra_args.append("--" + a)
     
     def compile(self, module_path: str, *, save_to : str = None) -> str:
-        vmfb_path = save_to + "compiled_model.vmfb"
+        vmfb_path = os.path.join(save_to, "compiled_model.vmfb")
         arg_string = f"--iree-hal-target-backends={self.hal_target_backend} "
         for arg in self.extra_args:
             arg_string += arg
             arg_string += " "
-        command_error_dump = os.path.join(save_to, "detail/compilation.detail.log")
-        commands_log = os.path.join(save_to, "commands/compilation.commands.log")
-        script = f"iree-compile {module_path} {arg_string}-o {vmfb_path} 2> {command_error_dump}"
+        command_error_dump = os.path.join(save_to, "detail", "compilation.detail.log")
+        commands_log = os.path.join(save_to, "commands", "compilation.commands.log")
+        script = f"iree-compile {module_path} {arg_string}-o {vmfb_path} 1> {command_error_dump} 2>&1"
         with open(commands_log, "w") as file:
             file.write(script) 
         # remove old vmfb if it exists
--- a/alt_e2eshark/e2e_testing/backends.py
+++ b/alt_e2eshark/e2e_testing/backends.py
@@ -6,9 +6,11 @@
 import abc
 import onnxruntime as ort
 from typing import TypeVar, List
-from e2e_testing.storage import TestTensors
+from e2e_testing.storage import TestTensors, get_shape_string
 from e2e_testing.framework import CompiledOutput, ModelArtifact
 from onnx import ModelProto
+import os
+from pathlib import Path
 
 Invoker = TypeVar("Invoker")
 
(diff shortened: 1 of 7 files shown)


zjgarvey:
Going to add a few nit commits here in a moment:
1. store detail logs in a sub-folder to avoid cluttering the log directory. 
2. fix passing the test name through func_name in the config.run method.

alt_e2eshark/e2e_testing/backends.py
```diff
@@ -94,6 +96,55 @@ def func(x):
...
+                    self.extra_args.append("--" + a)
+    
+    def compile(self, module_path: str, *, save_to : str = None) -> str:
+        vmfb_path = save_to + "compiled_model.vmfb"
+        arg_string = f"--iree-hal-target-backends={self.hal_target_backend} "
+        for arg in self.extra_args:
+            arg_string += arg
+            arg_string += " "
+        command_error_dump = os.path.join(save_to, "detail/compilation.detail.log")
+        commands_log = os.path.join(save_to, "commands/compilation.commands.log")
```
monorimet, line 119: change to os.path.join for the whole thing instead of unix style paths here -- just for windows compatibility

alt_e2eshark/e2e_testing/test_configs/onnxconfig.py
```diff
@@ -112,3 +114,132 @@ def compile(self, mlir_module: Module, *, save_to: str = None) -> CompiledArtifa
...
+        if not save_to:
+            raise ValueError("CLOnnxTestConfig requires saving artifacts")
+        # setup a detail subdirectory
+        os.makedirs(os.path.join(save_to, "detail"), exist_ok=True)
+        # setup a commands subdirectory
+        os.makedirs(os.path.join(save_to, "commands"), exist_ok=True)
+        # set file paths
+        mlir_file = save_to + "model.torch_onnx.mlir"
+        detail_log = os.path.join(save_to, "detail/import_model.detail.log")
+        commands_log = os.path.join(save_to, "commands/import_model.commands.log")
```
monorimet, line 142: same

alt_e2eshark/e2e_testing/backends.py
ScottTodd: You can also use `pathlib.Path` (https://docs.python.org/3/library/pathlib.html). I find that much nicer to use than `os` for path manipulation.

zjgarvey: Yeah, I really like pathlib, personally. I have half a mind to convert all of this testsuite to using `pathlib.Path` instead of `str`, but sometimes I do actually need to have a string, and I start with a string, so `os.path.join` seems easier than `str(Path(thing).joinpath(otherthing))`.

zjgarvey: I wonder if it's doubly bad that I use pathlib *sometimes*. Especially when getting parent directories, names, etc.

monorimet (approved):
LGTM. Can probably switch some of the CL script running to subprocess.run() calls as a follow-up.

alt_e2eshark/e2e_testing/test_configs/onnxconfig.py
```diff
@@ -112,3 +114,132 @@ def compile(self, mlir_module: Module, *, save_to: str = None) -> CompiledArtifa
...
+        # setup a detail subdirectory
+        os.makedirs(os.path.join(save_to, "detail"), exist_ok=True)
+        # setup a commands subdirectory
+        os.makedirs(os.path.join(save_to, "commands"), exist_ok=True)
+        # set file paths
+        mlir_file = os.path.join(save_to, "model.torch_onnx.mlir")
+        detail_log = os.path.join(save_to, "detail", "import_model.detail.log")
+        commands_log = os.path.join(save_to, "commands", "import_model.commands.log")
+        # get a command line script
+        script = "python -m torch_mlir.tools.import_onnx "
```
saienduri, line 144: Hey, can we possibly do this? (I haven't tried it)
image
Might help us avoid some python path stuff

zjgarvey: Ah, good point. I've been using the setup for a local torch-mlir build. Let me test out with a pip installed torch-mlir

zjgarvey: Running from python seems to work fine with a pip installed torch-mlir, too. When using a local build, you have to setup your pythonpath anyway, so I think it's fine. I've got a script for saving environment variables (e.g., pythonpath) to a `.env` file in the utils directory for people using local builds.

Also, I am unable to use the `torch-mlir-import-onnx` command-line tool from my local build without adding it to my path. Even worse, `torch-mlir-import-onnx` from a pip install actually throws an error:

```
  File "/home/zjgar/code/SHARK-TestSuite/alt_e2eshark/pipped.venv/bin/torch-mlir-import-onnx", line 5, in <module>
    from torch_mlir.tools.import_onnx import _cli_main
ImportError: cannot import name '_cli_main' from 'torch_mlir.tools.import_onnx' (unknown location)
```

Might be a good idea to file an issue for this...

zjgarvey: I'm going to merge this PR in a little bit, unless you want some time to review it further. Let me know, @saienduri .

saienduri: Looks good to me Zach, thanks for this!

saienduri: So, from the CI point of view, does it need to set any python environment variables or does the pip installed wheel work fine?

zjgarvey: Pip should work great!

zjgarvey: Just need to set the CACHE_DIR environment variable and the AZ_PRIVATE_CONNECTION (or whatever that var is called)

PR #345, opened 2024-09-19
[alt] Misc updates

zjgarvey:
This PR includes quite a few changes:

1. include a status notification for current stage
2. remove optimizations from the ort session used to get model signature
3. sort the test_list before running tests
4. modify inputs for an migraphx model
5. fix an issue that occurs when loading scalar tensors from .bin files
6. add names to un-named nodes when truncating onnx models
7. return to printing the whole exception on a stage failure
8. don't reproduce detail log contents in top level error messages.


saienduri (approved):
LGTM, thanks

PR #356, opened 2024-09-30
[alt] remove default iree-compile extra args from onnx-iree mode

zjgarvey:
Always using `iree-input-demote-i64-to-i32` was causing numerics problems since it wasn't loading i64 inputs correctly. Additionally the flags for specifying the chip for `hip` is out of date, and should be removed. Users should specify `-ica "iree-hip-target=gfx942" "other-arg" "other-arg" ...` if compiling for Mi300.


alt_e2eshark/e2e_testing/backends.py
```diff
@@ -34,36 +34,17 @@ class SimpleIREEBackend(BackendBase):
     '''This backend uses iree to compile and run MLIR modules for a specified hal_target_backend'''
     def __init__(self, *, device="local-task", hal_target_backend="llvm-cpu", extra_args : List[str] = None):
         self.device = device
+        if hal_target_backend == "hip":
+            print("IREE compiler python bindings do not currently support the change to using iree-hal-target-device=hip. Defaulting to depreciated iree-hal-target-backends=rocm")
+            hal_target_backend = "rocm"
```
ScottTodd, line 39: Got more context for this? Something doesn't seem right here...

also typo: "deprecated" not "depreciated"

alt_e2eshark/e2e_testing/backends.py
```diff
@@ -111,7 +95,7 @@ def __init__(self, *, device="local-task", hal_target_backend="llvm-cpu", extra_
     
     def compile(self, module_path: str, *, save_to : str = None) -> str:
         vmfb_path = os.path.join(save_to, "compiled_model.vmfb")
-        arg_string = f"--iree-hal-target-backends={self.hal_target_backend} "
+        arg_string = f"--iree-hal-target-device={self.hal_target_device} "
```
ScottTodd, line 98: Note that this `iree-hal-target-device` flag is quite new, and we aren't even using it in-tree in iree-org/iree for many tests yet.

alt_e2eshark/e2e_testing/backends.py
zjgarvey: I'm looking at CompilerOptions and, for example, <https://github.com/iree-org/iree/blob/35dafeeb73ad7ace5b81b6250c3c29f6809ee2b4/compiler/bindings/python/iree/compiler/tools/core.py#L187>. This still uses `--iree-hal-target-backends` rather than `-iree-hal-target-device`. 

Will change the typo, thanks.

zjgarvey:
@ScottTodd Should I revert the change to the cl config to use `--iree-hal-target-backends` instead of `--iree-hal-target-device`?

ScottTodd:
> @ScottTodd Should I revert the change to the cl config to use `--iree-hal-target-backends` instead of `--iree-hal-target-device`?

That's probably safer for now. Since the new flag hasn't been fully adopted in the IREE repo and doesn't have much mileage on it, it may be buggy. That's going to be easier to reason about upstream first, and that sort of flag change can also be made independently from the rest of the cleanup here.

vinayakdsci (approved):
LGTM, thanks! But let's wait for an approval from @ScottTodd too.

alt_e2eshark/e2e_testing/backends.py
```diff
@@ -160,12 +138,6 @@ def __init__(self, *, device="local-task", hal_target_device="llvm-cpu", extra_a
             self.extra_args = [
                 "--iree-hip-target=gfx90a",
             ]
```
ScottTodd, line 140: FYI, we've been pulling this from an environment variable in other tests. See https://github.com/nod-ai/SHARK-Platform/pull/373#discussion_r1822904806 for some discussion

ScottTodd (approved):

alt_e2eshark/e2e_testing/backends.py
zjgarvey: Yeah, that's a good idea. It's a bit verbose to need to write out all the flags repeatedly.

PR #357, opened 2024-10-01
Fix bert nlp models failing input generation by overriding `token_type_ids` to limit its range to [0,2) and some refactoring

renxida:
Note: this causes each nlp model to create an additional onnx inference session. A larger scale refactoring may be required to fix this.

Testing: I ran this on local with a bert model and was able to get to a numeric mismatch.

Full list of changes:

1. Refactor input generation for ONNX models:
   - Split `get_node_shape_from_dim_param_dict()` out of 
     `generate_input_from_node()` for better modularity.
   - Improve docstrings for better function descriptions.

2. Enhance type annotations:
   - Add return type hints to `construct_inputs()` and 
     `get_sample_inputs_for_onnx_model()`.
   - Use `TestTensors` type consistently for input/output tensors.

3. Fix NLP(Bert)-specific input generation:
   - Override `construct_inputs()` in NLP models to handle 'token_type_ids'
     correctly.

### DIFF
--- a/alt_e2eshark/onnx_tests/models/nlp.py
+++ b/alt_e2eshark/onnx_tests/models/nlp.py
@@ -8,6 +8,12 @@ from pathlib import Path
 from ..helper_classes import AzureDownloadableModel
 from e2e_testing.registry import register_test
 from e2e_testing.storage import load_test_txt_file
+from e2e_testing.onnx_utils import get_node_shape_from_dim_param_dict
+from e2e_testing.storage import TestTensors
+import onnxruntime as ort
+import numpy
+from typing import Optional
+import os
 
 this_file = Path(__file__)
 lists_dir = (this_file.parent).joinpath("external_lists")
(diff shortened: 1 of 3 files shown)


alt_e2eshark/onnx_tests/models/nlp.py
```diff
@@ -28,6 +34,23 @@ def __init__(self, *args, **kwargs):
...
             self.dim_param_dict = dim_param_dict
 
+        def construct_inputs(self):
+            """Overrides the parent class method to construct sample inputs with the correct dimensions."""
+            default_inputs = super().construct_inputs()
+
+            tensors = list(default_inputs.data)
+
+            self.update_sess_options()
+            session = ort.InferenceSession(self.model, self.sess_options)
```
zjgarvey, line 44: I don't really want to load an inference session again for this. Can we store a list of corresponding input names when running construct inputs the first time (maybe in the test info class)? This might be useful for other purposes, too.

zjgarvey (commented):
It's quite time-expensive to be making several of these inference sessions, and you might be able to get around this by finding a way to store input node names at some stage of the process. Either when constructing inputs, or maybe just any place where an onnx model is getting loaded.

alt_e2eshark/onnx_tests/models/nlp.py
zjgarvey: Let's merge this anyway for now.

zjgarvey (approved):
Let's merge for now. We can try to get rid of the extra ORT session in a future patch.

PR #358, opened 2024-10-01
[alt] don't divide by zero in report summary

zjgarvey:


alt_e2eshark/utils/report.py
```diff
@@ -76,7 +80,7 @@ def get_exit_status_string(counts: Dict[str, int], total) -> str:
     for key, value in counts.items():
         if key == "PASS":
             continue
-        results_str += f"| {key} | {value} | {round((value/total)*100, 1)}% |\n"
+        results_str += f"| {key} | {value} | {round(safe_div(value, total)*100, 1)}% |\n"
```
ScottTodd, line 83: Could also do something like
```python
percent_of_total_str = total == 0 ? "N/A" : f"{round((value / total)*100, 1)}%"
results_str += f"| {key} | {value} | {percent_of_total_str} |\n"
```

ScottTodd (approved):

alt_e2eshark/utils/report.py
zjgarvey: Yeah, I thought about using ternaries but it ended up cluttering the code a bit more. `N/A` is probably more correct, but I think `0.0%` is reasonable for reporting 0 passes out of 0 runs. 

I really need to consider using a table package rather than generating markdown strings haha.

PR #366, opened 2024-10-10
[alt][docs] Address non-GPU systems

marbre:


zjgarvey (approved):
I did not know you could do that with a pytorch install. 

Thanks for looking into this and posting an update Marius.

marbre:
> I did not know you could do that with a pytorch install.
> 
> Thanks for looking into this and posting an update Marius.

I didn't knew before either ;) Thanks for the quick review!

PR #367, opened 2024-10-11
make some adjustments to the torch-onnx to linalg pass pipeline

zjgarvey:
I would normally add all of the torch related stuff to the "onnx-to-torch-backend-pipeline" but there is a common issue with slice ops having an int that's too big only when saving/loading an mlir file at the torch level. See <https://github.com/llvm/torch-mlir/issues/3621>.

Also adds a few models to the basic opt list, since they would need a whole 'nother cycle of shape inference and scalarization to clean themselves up. Good to know that it is possible to clean these up at the torch level though!


kumardeepakamd (approved):
Looks good. Can go ahead and merge.

PR #370, opened 2024-10-15
Add tests for onnx op ConvTranspose support for autopadding.

guacamoleo:
Add tests for onnx op ConvTranspose support for autopadding (https://github.com/llvm/torch-mlir/pull/3797).

- Original onnx autopad tests required asymmetric padding (which still isn't supported); therefore new valid tests were required.
- Adds tests for explicit padding and auto-pading with and without output_shape specified.


zjgarvey (changes requested):
I'd add two more tests for "SAME_LOWER" and "VALID" to see if everything works out in those cases.

guacamoleo:
Thanks for the quick feedback! I added tests for SAME_LOWER and VALID.

zjgarvey (approved):
These look like good test cases, thanks!

PR #374, opened 2024-10-21
Fix some quirks in get shape string, fix saving an int32 input

zjgarvey:
Following up on some work by @AmosLewis, I noticed another bug with i32 inputs.


saienduri (approved):
lgtm

AmosLewis (approved):
LGTM

PR #376, opened 2024-10-23
Add ability to gather and analyse some model metadata

zjgarvey:
## Usage:

When running `python run.py <other-args> --get-metadata`, this will save a dictionary with the model size and op frequencies to the log directory.

After a run, you can use `python utils/find_duplicate_models.py` to save or print a json dump of redundant models.

### Options:
- "-s" "--simplified" will only return the list of model names (doesn't include the corresponding metadata).
- "-o" "--output" allows specifying the name of a json file you want to save the result to.
- "-r" "--rundirectory" allows specifying a different run directory to search, if `run.py` was run with a non-default run directory arg. 

## Sample:

I saved the tests below to a file called `sample.txt`.

```
add_test
model--bart-base-booksum--KamilAin
model--bart-base-cnn--ainize
model--bart-base-few-shot-k-1024-finetuned-squad-seed-2--anas-awadalla
model--bart-base-few-shot-k-1024-finetuned-squad-seed-4--anas-awadalla
```

With a clean `test-run` directory, I ran 
```shell
python run.py --testsfile=sample.txt --stages "setup" --get-metadata
```

The result of running 
```shell
python utils/find_duplicate_models.py -s
```
was:

```json
[
    [
        "model--bart-base-booksum--KamilAin",
        "model--bart-base-cnn--ainize"
    ],
    [
        "model--bart-base-few-shot-k-1024-finetuned-squad-seed-4--anas-awadalla",
        "model--bart-base-few-shot-k-1024-finetuned-squad-seed-2--anas-awadalla"
    ]
]
```

and without the `-s` arg, it includes the metadata for each grouping:

```json
[
    {
        "models": [
            "model--bart-base-booksum--KamilAin",
            "model--bart-base-cnn--ainize"
        ],
        "shared_metadata": {
            "model_size": 712772272,
            "op_frequency": {
                "Add": 227,
                "Cast": 13,
                "Concat": 188,
                "Constant": 886,
                "ConstantOfShape": 6,
                "Div": 44,
                "Equal": 5,
                "Erf": 12,
                "Expand": 5,
                "Gather": 64,
                "Less": 1,
                "MatMul": 133,
                "Mul": 99,
                "Pow": 32,
                "Range": 3,
                "ReduceMean": 64,
                "Reshape": 187,
                "Shape": 67,
                "Slice": 2,
                "Softmax": 18,
                "Sqrt": 32,
                "Squeeze": 2,
                "Sub": 35,
                "Transpose": 90,
                "Unsqueeze": 325,
[...]

PR #379, opened 2024-10-28
removing model which are not valid and duplicates

pdhirajkumarprasad:
removing models from list which are not valid based on https://github.com/nod-ai/SHARK-ModelDev/issues/812 and remove duplicates using tool introduced here: https://github.com/nod-ai/SHARK-TestSuite/pull/376


saienduri:
LGTM, let's wait for @zjgarvey before landing

zjgarvey (commented):
I'd like to retain the original shard lists somewhere. I'm not sure if Kumar or others want to still have numbers reported based on the full list with redundant tests.

 Do we also need to redistribute across shards?

pdhirajkumarprasad:
> I'd like to retain the original shard lists somewhere. I'm not sure if Kumar or others want to still have numbers reported based on the full list with redundant tests.

--I can keep the original list with extension _old but our run should be with the unique model list and it's fine if that lead to lesser no. of model but that will be unique list  

> Do we also need to redistribute across shards?

--Yes, I can do that

pdhirajkumarprasad:
closing this pull request as the way DUP model was found is not correct and it will reduce the coverage of ops. Once we have correct list, will create the pull request

PR #388, opened 2024-11-06
Pulling in Alt E2ESHARK CI changes

saienduri:
This pulls in the required changes to run nightly CI for alt e2eshark.

Few things to iterate on here:

- Figure out a way to run a loop/matrix on a step so we don't have a bunch of download commands in push artifact step
- Figure out how to not fail the download artifact step if an artifact for a certain shard is not available
- Pull the merge report script calls, creating directories, copy pasting artifacts to eventually push into the reports repo into a python script that won't fail if certain shards fail and are not available

Just merging this in so we have a nightly going that doesn't depend on manual triggers. We can iterate on it. Has been running fine for the last week.

### DIFF
--- a/alt_e2eshark/e2e_testing/backends.py
+++ b/alt_e2eshark/e2e_testing/backends.py
@@ -88,11 +88,11 @@ class CLIREEBackend(BackendBase):
                 else:
                     self.extra_args.append("--" + a)
         if hal_target_backend == "rocm":
-            self.extra_args = [
+            self.extra_args += [
                 f"--iree-hip-target={self.target_chip}",
             ]
         if hal_target_backend == "llvm-cpu":
-            self.extra_args = [
+            self.extra_args += [
                 "--iree-llvmcpu-target-cpu=host",
             ]
     
--- a/alt_e2eshark/e2e_testing/backends.py
+++ b/alt_e2eshark/e2e_testing/backends.py
@@ -87,11 +87,11 @@ class CLIREEBackend(BackendBase):
                     self.extra_args.append(a)
                 else:
                     self.extra_args.append("--" + a)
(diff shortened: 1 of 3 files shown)


alt_e2eshark/run.py
```diff
@@ -331,6 +331,12 @@ def _get_argparse():
         default="llvm-cpu",
         help="specifies the iree-hal-target-device / iree-hal-target-backends for compile phase",
     )
+    parser.add_argument(
+        "-t",
```
zjgarvey, line 335: This is already used by another arg.

alt_e2eshark/e2e_testing/backends.py
```diff
@@ -76,16 +76,25 @@ def func(x):
...
         self.hal_target_backend = hal_target_backend
+        self.target_chip = target_chip
         self.extra_args = []
         if extra_args:
             for a in extra_args:
                 if a[0:2] == "--":
                     self.extra_args.append(a)
                 else:
                     self.extra_args.append("--" + a)
+        elif hal_target_backend == "rocm":
```
zjgarvey, line 90: Maybe this should just be an "if" and instead of overwriting the extra args, it appends to it? Same w/ below

zjgarvey (changes requested):

alt_e2eshark/e2e_testing/backends.py
saienduri: Good point missed it in copy pasting the blocks from elsewhere

alt_e2eshark/run.py
saienduri: Changed to -c

zjgarvey (approved):
Thanks!

PR #393, opened 2024-11-13
Support downloads from ONNX Model Zoo

vinayakdsci:
Adds preliminary support for pulling and running models from the ONNX Model Zoo in SHARK-TestSuite.

Co-authored-by: zjgarvey <47986913+zjgarvey@users.noreply.github.com>


alt_e2eshark/onnx_tests/models/external_lists/onnx_model_zoo.txt
pdhirajkumarprasad, line 1: Please split this into 4 list so that we have ~500 models in each list and that way, it will reduce overall CI time

pdhirajkumarprasad (changes requested):
Split the onnx_model_zoo.txt into list of 4 files and do the change accordingly in helper_classes.py

zjgarvey (commented):
I think we want at most around 500 models per shard, so it would be good to break up the computer vision models into multiple files. Maybe we can do like 4 files with 480ish models each.

alt_e2eshark/e2e_testing/onnx_utils.py
```diff
@@ -29,6 +29,12 @@ def dtype_from_ort_node(node):
 
 def get_node_shape_from_dim_param_dict(node: onnxruntime.capi.onnxruntime_pybind11_state.NodeArg, dim_param_dict: Optional[dict[str, int]] = None):
     """Get the shape of a node, replacing any string dims with values from a dim_param_dict"""
+    if dim_param_dict and node.name in dim_param_dict.keys():
+        # Particularly used for models from ONNX Model Zoo,
+        # the dim_param_dict here should contain the shapes
+        # of the input nodes.
+        return dim_param_dict[node.name]
+
```
zjgarvey, line 37: The node name shouldn't be a key to the `dim_param_dict`. The `dim_param_dict` should take the string name of the dim parameter itself as a key. We could plumb through an alternative association of `input_node_name -> [static_shape, dtype]`, but this would be a separate dictionary.

alt_e2eshark/run.py
```diff
@@ -223,7 +223,10 @@ def run_tests(
             curr_stage = "construct_inputs"
             if curr_stage in stages:
                 notify_stage()
-                if load_inputs:
+                model_zoo_inputs = False
+                if os.path.exists(log_dir + "/input_0.pb"):
+                    model_zoo_inputs = True
+                if load_inputs or model_zoo_inputs:
```
zjgarvey, line 229: Instead of doing this, override the `construct_inputs` method for your `OnnxModelInfo` subclass to `return self.load_inputs(log_dir)`.

alt_e2eshark/onnx_tests/models/onnx_zoo_models.py
```diff
@@ -0,0 +1,21 @@
...
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from e2e_testing.registry import register_test
+from ..helper_classes import (
+    OnnxModelZooDownloadableModel,
+    onnx_zoo_validated,
+    onnx_zoo_non_validated,
```
zjgarvey, line 11: I'd prefer to have these lists in this particular file, rather than in `helper_classes.py`.

I realize that you use these lists to determine some control flow in your model info class, but it may be better to have a class attribute bool `is_validated` that performs the same task. 

You could add an arg at the beginning of the `__init__` method, and then define in this file 

```python
validated_constructor =  lambda args, kwargs : OnnxModelZooDownloadableModel(True, *args, **kwargs)
non_validated_constructor =  lambda args, kwargs : OnnxModelZooDownloadableModel(False, *args, **kwargs)
```

zjgarvey (changes requested):
This mostly looks good to me, but I'd like a few changes before merging.

alt_e2eshark/onnx_tests/models/onnx_zoo_models.py
vinayakdsci: @zjgarvey this is actually rather difficult to do, as the constructor for the TestConfig only allows me to pass in the model_name and the model_constructor. What I have done for now is that moved these lists and the `model_path_map` to this file, and simply imported `model_path_map` in `helper_classes.py`.

zjgarvey (approved):

PR #398, opened 2024-11-22
Fix nlp model input attention_mask issue

AmosLewis:
model--bsc-bio-ehr-es-pharmaconer--PlanTL-GOB-ES pass the test with this fix.

The attention_mask input has to limit to 0/1


zjgarvey (approved):
Nice, this seems like an easy fix. Let's see how the run looks tonight with this change.

PR #407, opened 2024-12-04
Added model to basic opt list to make it work

pdhirajkumarprasad:
After adding them in opt list, it's passing numerics


vivekkhandelwal1 (approved):
LGTM

PR #409, opened 2024-12-09
Support downloads from Huggingface Model Hub

vinayakdsci:
[WIP] Adds support to download and run models from the Huggingface Model Hub.

Some work is still needed in input construction and caching for different model classes. All models in the list currently pass setup.


alt_e2eshark/onnx_tests/helper_classes.py
```diff
@@ -23,6 +25,50 @@
...
+        self.task = task_name
+        super().__init__(name, onnx_model_path, opset_version)
+
+    def export_model(self):
+        model_dir = str(Path(self.model).parent)
+        # TODO: Use Python API instead of CLI?
+        opt_cli_command = f"optimum-cli export onnx --model {self.model_repo_path} --task {self.task} --monolith --framework pt {model_dir}"
+        cli_output = subprocess.run(opt_cli_command.split(' '), capture_output=True)
+        # TODO: Find a cleaner way to exit than an assert.
+        assert cli_output.returncode == 0, f"Failed to run optimum-cli:\n{cli_output.stderr.decode()}"
```
zjgarvey, line 43: I don't want to crash the test runner if this fails. Let's raise an exception instead. E.g. :
```suggestion
        if cli_output.returncode == 0:
           raise RuntimeError(f"Failed to run optimum-cli:\n{cli_output.stderr.decode()}")
```

alt_e2eshark/onnx_tests/helper_classes.py
```diff
@@ -23,6 +25,50 @@
...
+    def __init__(self, full_model_path, task_name, name, onnx_model_path):
+        opset_version = 21
+        self.model_repo_path = full_model_path.replace("hf_", "")
+        self.task = task_name
+        super().__init__(name, onnx_model_path, opset_version)
+
+    def export_model(self):
+        model_dir = str(Path(self.model).parent)
+        # TODO: Use Python API instead of CLI?
+        opt_cli_command = f"optimum-cli export onnx --model {self.model_repo_path} --task {self.task} --monolith --framework pt {model_dir}"
```
zjgarvey, line 40: IIRC, optimum doesn't support exporting some model types like FLUX. 

I'm also wondering if we should avoid using OnnxModelInfo, and instead define a new ModelInfo base class for models originally in pytorch, which we could run natively in pytorch for the gold output. 

The reason I'd probably prefer this approach is that for the large models we eventually want to export with initializers converted to inputs, we won't really be able to run those with ORT. 

Let me think about this a bit more. It might be fine to leave this as-is for the medium sized models and do something else for the extremely large models.

vinayakdsci: That would be easy to do. I wrote it this way so that we can have a custom registry for very large models or the ones that optimum doesn't support, and override this function for those few models. Having optimum for the bigger set allows for easier implementation, and debugging IMO.

zjgarvey (approved):
Lets get this in and do a run to see what needs to be changed. Looks pretty good to me right now. 

Noticing some similarities in pathing issues with onnx model zoo models, I'd also like to invest some time in being able to store test name to test metadata dictionaries as json files and be able to use --testsfile with jsons as well as txt files.

vinayakdsci:
Thanks @zjgarvey! I'll go ahead and merge this. Using json to store metadata sounds neat, we can discuss it and I can open a PR for that too.

PR #411, opened 2024-12-12
Add Flux_1.dev transformer and text encoder tests

zjgarvey:
The most significant infrastructure change is with some reworks to OnnxModelInfo:

1. `OnnxModelInfo` now has an attribute for `ExtraOptions`, which holds test-specific options for later testing stages beyond setup. This was required to allow running with external params, passing compile flags, and setting custom mlir import options.
2. All attributes configurable via `update_*` methods in `OnnxModelInfo` are updated in the init method immediately. This required reworking a few helper classes which relied on calling these at a specific time. 

Another infrastructure change is through the migration to using `iree-import-onnx` rather than torch-mlir's onnx importer. This is to allow externalizing parameters. 

Adds the following three tests:

- `flux_1_dev_transformer`
- `flux_1_dev_clip`
- `flux_1_dev_t5`

### DIFF
--- a/alt_e2eshark/e2e_testing/backends.py
+++ b/alt_e2eshark/e2e_testing/backends.py
@@ -7,7 +7,7 @@ import abc
 import onnxruntime as ort
 from typing import TypeVar, List
 from e2e_testing.storage import TestTensors, get_shape_string
-from e2e_testing.framework import CompiledOutput, ModelArtifact
+from e2e_testing.framework import CompiledOutput, ModelArtifact, CompilerOptions, RuntimeOptions
 from onnx import ModelProto
 import os
 from pathlib import Path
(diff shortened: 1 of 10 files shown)


alt_e2eshark/e2e_testing/backends.py
```diff
@@ -18,18 +18,23 @@
...
 
 
 from iree import compiler as ireec
 from iree import runtime as ireert
 
 
+def flag(arg : str) -> str:
+    if arg.startswith("--"):
+        return arg
+    else:
```
vinayakdsci, line 36: I think you can leave out this else, and simply return.
```suggestion
    return f'--{arg}'
```

Nit: since arg is supposed to be an str, slicing could be a better option here given that arg is not too long in length.

alt_e2eshark/e2e_testing/backends.py
```diff
@@ -96,12 +101,15 @@ def __init__(self, *, device="local-task", hal_target_backend="llvm-cpu", target
...
+        if self.hal_target_backend in extra_options.backend_specific_flags.keys():
+            test_specific_args += extra_options.backend_specific_flags[self.hal_target_backend]
+        compile_args = self.extra_args + [flag(arg) for arg in test_specific_args]
         vmfb_path = os.path.join(save_to, "compiled_model.vmfb")
         arg_string = f"--iree-hal-target-backends={self.hal_target_backend} "
-        for arg in self.extra_args:
-            arg_string += arg
-            arg_string += " "
+        for arg in compile_args:
+            arg_string += f'{arg} '
```
vinayakdsci, line 112: I know a neat one-liner to replace the for loop :)
```suggestion
            arg_string += " ".join([*compile_args, ""])
```
The empty string should append a whitespace to `arg_string`.

vinayakdsci (approved):
LGTM, thanks!

PR #412, opened 2024-12-12
Fix setup failures for HF Models

vinayakdsci:
Fixes a number of setup failures in HF Model support.

The `HfDownloadableModel` now respects the cache directory. and multiple choice models have better input generation.

Also fixes input mismatch between tokenization and the ONNX models in Deberta and Roberta models.


pdhirajkumarprasad (approved):
approved

PR #414, opened 2024-12-12
Add `tf-keras` to required packages

vinayakdsci:
Adds the package `tf-keras` to `base_requirements.txt` for backwards compatibility in HuggingFace models that require `tensorflow` during export.

### DIFF
--- a/alt_e2eshark/base_requirements.txt
+++ b/alt_e2eshark/base_requirements.txt
@@ -16,4 +16,5 @@ azure-storage-blob
 compressed_tensors
 fugashi
 tensorflow
+tf-keras
 timm


pdhirajkumarprasad (approved):

zjgarvey:
How do you feel about having a separate requirements file for hf model dependencies?

For people not running hf models, if these packages are large, I'd prefer to not have to install them.

vinayakdsci:
> How do you feel about having a separate requirements file for hf model dependencies?
> 
> For people not running hf models, if these packages are large, I'd prefer to not have to install them.

@zjgarvey sounds good to me. So you wish me to create a patch for that?

PR #418, opened 2024-12-17
Delete iree_tests/sharktank/.

ScottTodd:
These tests are redundant with newer test suites here:

* https://github.com/iree-org/iree/tree/main/experimental/regression_suite/shark-test-suite-models (sd3 and sdxl)
  * Note that these tests use files from https://huggingface.co/amd-shark/sdxl-quant-models, while the tests there use Azure (sharkpublic)
* https://github.com/iree-org/iree-test-suites/tree/main/sharktank_models (llama)
  * Note that llama 3 tested there is different from the [openlm-research/open_llama_3b_v2](https://huggingface.co/openlm-research/open_llama_3b_v2) model tested here, but we should move tests to iree-test-suites anyways: https://github.com/iree-org/iree-test-suites/issues/49


.github/workflows/test_iree.yml
```diff
@@ -77,15 +77,13 @@ jobs:
...
           source ${VENV_DIR}/bin/activate
           python3 iree_tests/download_remote_files.py --root-dir iree_tests/pytorch/models
-          python3 iree_tests/download_remote_files.py --root-dir iree_tests/sharktank
 
       - name: "Running real weight model tests"
         if: "matrix.models-config-file != '' && !cancelled()"
         run: |
           source ${VENV_DIR}/bin/activate
           pytest \
             iree_tests/pytorch/models \
```
ScottTodd, line 86: I can start porting these to https://github.com/iree-org/iree-test-suites as well. Just opt-125M and resnet50 in https://github.com/nod-ai/SHARK-TestSuite/tree/main/iree_tests/pytorch/models right now since no one helped port from e2eshark x_x, we wouldn't lose much by dropping that coverage without a replacement.

saienduri:
Cool, this makes sense. Are you going to be doing the follow up PR to remove this testing in iree also?

ScottTodd:
> Cool, this makes sense. Are you going to be doing the follow up PR to remove this testing/updating pin in iree also? I actually don't think we even run any other SHARK-TestSuite tests in iree, so we can probably fully remove shark test suite from iree testing

Yeah these run in IREE: https://github.com/iree-org/iree/blob/fb4d09470dc4be674de810dbfbf2d3764e2970ba/.github/workflows/pkgci_regression_test.yml#L83-L117. That `test_models` matrix.
```
          pytest \
            SHARK-TestSuite/iree_tests/pytorch/models \
            SHARK-TestSuite/iree_tests/sharktank \
```

I don't mind dropping the two pytorch/models/ tests at the same time. The coverage provided there is small at the moment, and I'm reasonably close to having a replacement ready in https://github.com/iree-org/iree-test-suites/issues/63.

saienduri:
Makes sense, sgtm

saienduri (approved):

PR #423, opened 2025-01-03
Cleanup issues with a few helper classes

zjgarvey:
This patch

1. fixes setup for zoo models to align with `update_*` methods getting called in `__init__`
2. updates opset version for hf models on import by default
3. moves an import of optimum into the HF model info class `__init__` method
4. applies black formatting

### DIFF
--- a/alt_e2eshark/onnx_tests/helper_classes.py
+++ b/alt_e2eshark/onnx_tests/helper_classes.py
@@ -37,19 +37,15 @@ class HfDownloadableModel(OnnxModelInfo):
     """This class should be used to download models from Huggingface model hub."""
 
     def __init__(self, full_model_path, task_name, name, onnx_model_path):
-        # Checking if CACHE_DIR is set here will allow us to redefine
-        # HF_HOME and HUGGINGFACE_HUB_CACHE without requiring them to
-        # be set at shell level.
         parent_cache_dir = os.getenv("CACHE_DIR")
         if not parent_cache_dir:
             raise RuntimeError(
                 "Please specify a cache directory path in the CACHE_DIR environment variable "
                 "for storing large model files."
             )
-        os.environ["HF_HOME"] = "" if parent_cache_dir is None else parent_cache_dir
-        os.environ["HUGGINGFACE_HUB_CACHE"] = (
-            "" if parent_cache_dir is None else parent_cache_dir
-        )
+        # set HF cache explicitly to match provided CACHE_DIR
+        os.environ["HF_HOME"] = parent_cache_dir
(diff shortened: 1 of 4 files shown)


zjgarvey:
I've verified that this PR allows running non-cache required tests without setting CACHE_DIR (and doesn't warn on transformers or diffusers import). 

Onnx zoo tests also work with this change.

HF model tests also work with this change and no-longer save input files to `/test-run/<model-name>.0.bin`.

alt_e2eshark/onnx_tests/helper_classes.py
```diff
@@ -30,25 +30,21 @@
...
 class HfDownloadableModel(OnnxModelInfo):
     """This class should be used to download models from Huggingface model hub."""
 
     def __init__(self, full_model_path, task_name, name, onnx_model_path):
-
+        # Checking if CACHE_DIR is set here will allow us to redefine
+        # HF_HOME and HUGGINGFACE_HUB_CACHE without requiring them to
+        # be set at shell level.
+        parent_cache_dir = os.getenv("CACHE_DIR")
+        os.environ["HF_HOME"] = "" if parent_cache_dir is None else parent_cache_dir
```
vinayakdsci, line 44: I think we should raise an exception here if CACHE_DIR is not set before setting the env vars, just like for other classes. This is subtle :) If I do not set CACHE_DIR before running an HF model, the HF_HOME variable will still be set (to "", interpreted as PWD by bash), and then the exception will be raised.

Having the check before changing the variables would save a great deal of unpleasantness, and cryptic log messages informing that the cache has been moved.

zjgarvey: Yeah, that makes sense. It looks like we were also setting `parent_cache_dir` twice.

alt_e2eshark/onnx_tests/helper_classes.py
```diff
@@ -8,58 +8,76 @@
...
         if not parent_cache_dir:
             raise RuntimeError(
-                "Please specify a cache directory path in the CACHE_DIR environment variable for storing large model files."
+                "Please specify a cache directory path in the CACHE_DIR environment variable "
+                "for storing large model files."
+            )
+        os.environ["HF_HOME"] = "" if parent_cache_dir is None else parent_cache_dir
+        os.environ["HUGGINGFACE_HUB_CACHE"] = (
+            "" if parent_cache_dir is None else parent_cache_dir
+        )
```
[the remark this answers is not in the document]
zjgarvey: Yeah, good point.

vinayakdsci (approved):
LGTM, thanks for the patch!

PR #424, opened 2025-01-06
Make `--mode=cl-onnx-iree` default for `run.py` and refactor script running

zjgarvey:
This makes `--mode=cl-onnx-iree` the default mode since it is the most well-supported. A few code improvements are also made:

1. The flag `--iree-compile-args` now uses `-a` for shorthand, since `-ica` is not really valid.
2. Refactors the script running for `cl-onnx-iree` through a util function `run_command_and_log` in `logging_utils.py`. 
3. Script running now uses `subprocess`, which more intelligently catches errors with `ret.returncode` rather than the existence of stage-specific output files. Also checks for `&&` in the script to mitigate some security concerns with adding extra compile args.
4. reorganizes some imports for convenience and to allow case-by-case requirement dependency
5. Raises a warning if `torch-mlir-opt` is not found when trying to run with `preprocessing` (i.e., `--torchtolinalg`).


alt_e2eshark/e2e_testing/logging_utils.py
```diff
@@ -8,8 +8,35 @@
...
+    script = command if isinstance(command, str) else subprocess.list2cmdline(command)
+    # setup a commands subdirectory (if it doesn't exist)
+    commands_dir = os.path.join(save_to, "commands")
+    os.makedirs(commands_dir, exist_ok=True)
+    commands_log = os.path.join(commands_dir, f"{stage_name}.commands.log")
+    # log the command
+    with open(commands_log, "w") as file:
+        file.write(script)
+    # run the command
+    ret = subprocess.run(script, shell=True, capture_output=True)
```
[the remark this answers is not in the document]
zjgarvey: Uh, not sure honestly. I'll look into it.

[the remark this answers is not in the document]
zjgarvey: Nah, I don't think we need it for any of the commands being run. I'm not using stuff like `&>` or anything shell specific anymore. I'll test it out without `shell=True` and make sure everything works.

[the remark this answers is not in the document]
zjgarvey: Ah, nevermind. It doesn't work. Here is an import failure message from `add_test` :

```
Failed test at stage import_model with exception:
[Errno 2] No such file or directory: 'python -m iree.compiler.tools.import_onnx /home/zjgar/code/SHARK-TestSuite/alt_e2eshark/test-run/add_test/model.onnx --num-elements-threshold=100 --params-scope=model -o /home/zjgar/code/SHARK-TestSuite/alt_e2eshark/test-run/add_test/model.torch_onnx.mlir'
  File "/home/zjgar/code/SHARK-TestSuite/alt_e2eshark/run.py", line 206, in run_tests
    model_artifact, func_name = config.import_model(
                                ^^^^^^^^^^^^^^^^^^^^
  File "/home/zjgar/code/SHARK-TestSuite/alt_e2eshark/e2e_testing/test_configs/onnxconfig.py", line 174, in import_model
    run_command_and_log(import_command, save_to, "import_model")
  File "/home/zjgar/code/SHARK-TestSuite/alt_e2eshark/e2e_testing/logging_utils.py", line 27, in run_command_and_log
    ret = subprocess.run(script, capture_output=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.11/subprocess.py", line 1955, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'python -m iree.compiler.tools.import_onnx /home/zjgar/code/SHARK-TestSuite/alt_e2eshark/test-run/add_test/model.onnx --num-elements-threshold=100 --params-scope=model -o /home/zjgar/code/SHARK-TestSuite/alt_e2eshark/test-run/add_test/model.torch_onnx.mlir'
```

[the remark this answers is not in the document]
zjgarvey: Hmmm, this seems to be causing some issues, specifically with the flag `--pass-pipeline='builtin.module(....)'`, e.g. with the `--torchtolinalg` flag during preprocessing, or if extra compile args are passed to iree-compile, for example. 

If I want to run it with subprocess without `shell=True` the single quotes need to be removed. However, running the reproducing script without the single quotes causes an error.

This is turning out to be a bit of a headache to move the shell scripts away from being actual shell scripts (while also being actual shell scripts intended to be run in a shell when debugging individual stages)...

alt_e2eshark/e2e_testing/test_configs/onnxconfig.py
```diff
@@ -208,7 +208,7 @@ def run(self, artifact: str, inputs: TestTensors, *, func_name=None, extra_optio
         func = self.backend.load(artifact, func_name=func_name, extra_options=extra_options)
         command = func(inputs)
         num_outputs = len(self.tensor_info_dict[test_name][0])
-        command.extend([f"--output=@'{os.path.join(run_dir, f'output.{i}.bin')}'" for i in range(num_outputs)])
+        command.extend([f"--output=@{os.path.join(run_dir, f'output.{i}.bin')}" for i in range(num_outputs)])
```
zjgarvey, line 211: perhaps leave `'` for command line reproducers, but strip them for `subprocess.call`?

vinayakdsci: `subprocess.call` should be able to handle single quotes. The patch right now uses `subprocess.run`, which doesn't. Could you try changing the function? I strongly expect that to resolve this issue.

zjgarvey: It doesn't look like `subprocess.call` allows single quotes either. 

I've decided to move back to `shell=True`. It isn't any worse than using `os.system`, and we don't have external users supplying commands to servers or machines. 

To mitigate concern, I decided to at least check for `&&` in each script, just to be sure that we don't run any additional commands during compilation.

alt_e2eshark/e2e_testing/logging_utils.py
```diff
@@ -8,8 +8,44 @@
...
+    with open(commands_log, "w") as file:
+        file.write(script)
+    # Although we don't currently host the test suite on any machines which take external requests,
+    # we should still check for potentially dangrous scripts.
+    # A user can append to the compile command with the flag `--extra-compile-args`
+    # E.g. we should avoid examples like:
+    # `python run.py -t add_test -i "&& <malicious command>"`
+    assert (
+        script.count("&&") == 0
+    ), f"Script may be dangrous to run as it contains '&&'.\nSee {commands_log}"
```
ScottTodd, line 34: Interesting 🤔 

This isn't sufficient to catch all script injection attacks, so I wonder if it could instill a false sense of security. One obvious workaround is to use `||` instead of `&&` for the same type of attack.

For runs triggered by GitHub, at the point where someone can edit the test code, they already have arbitrary remote code execution or can remove the checks, so any checks like this are meaningless. Hardening would help if we added a website (or some other remote endpoint) that let you trigger runs with more broad permissions.

If we do want to harden, some suggestions:

* Sanitize input up front (still on the backend), then have these lower level functions operate on trusted input only
* Use an allowlist of arguments or symbols (maybe ban `&` and `|`, or only allow `a-zA-Z0-9-\s\.` and a few other characters that show up in CLI options
* Run commands within an isolated container or virtual machine that is only granted minimal access to files and other resources

alt_e2eshark/e2e_testing/logging_utils.py
```diff
@@ -8,8 +8,44 @@
...
+    # run the command
+    ret = subprocess.run(script, shell=True, capture_output=True)
+
+    # if an error occured, raise an exception
+    if ret.returncode != 0:
+        detail_dir = os.path.join(save_to, "detail")
```
ScottTodd, line 40: FYI I prefer to use pathlib over `os.path` since it saves a few characters for code like this and uses dedicated `Path` classes instead of `str`

Here the changes would be:
```diff
-detail_dir = os.path.join(save_to, "detail")
+detail_dir = save_to / "detail"
```

ScottTodd (approved):

alt_e2eshark/e2e_testing/logging_utils.py
zjgarvey: Yeah, I think checking for unusual characters in `--extra-compile-args` would be better. I also agree that, for now, there is no point in restricting `python run.py -i "malicious fake arg"` if `python run.py && malicious command` is already possible for people using the test suite. 

I think I'm going to simply remove the check here entirely for now. I don't see this test suite ever being used in something allowing remote requests.

alt_e2eshark/e2e_testing/logging_utils.py
zjgarvey: Yeah, `Path` also has a `write_text` method which simplifies the code a bit.

I think a subsequent patch could go through elsewhere and clean up `os.path` stuff that can be converted to `pathlib`.

PR #425, opened 2025-01-06
Delete iree_tests folder.

ScottTodd:
The iree_tests folder has been migrated to a new repository at https://github.com/iree-org/iree-test-suites, see https://github.com/iree-org/iree-test-suites/issues/49. The only tests remaining here have been marked XFAIL since https://github.com/iree-org/iree/issues/17344 because the files needed to be regenerated:
```json
      "pytorch/models/opt-125M",
      "pytorch/models/resnet50",
    ],
```

See also https://github.com/iree-org/iree/pull/19614, which stops cloning this repository and running these tests in https://github.com/iree-org/iree.


saienduri (approved):
Thanks!

PR #426, opened 2025-01-07
Add llama 3.1 8b test

jinchen62:


zjgarvey:
It took a while, but I think I got a handle on the issue with input generation (and a few other issues that subsequently got uncovered). Please take a look at <https://github.com/jinchen62/SHARK-TestSuite/pull/1> when you get the chance.

PR #436, opened 2025-01-20
Add onnx.NonMaxSuppression

lpy:
This patch adds one basic test that passes in identical boxes, and one test that passes in boxes that use center point.

The center_point_box=1 support is added in llvm/torch-mlir#3976

### DIFF
--- a/alt_e2eshark/onnx_tests/operators/non_max_suppression.py
+++ b/alt_e2eshark/onnx_tests/operators/non_max_suppression.py
@@ -1,0 +1,112 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+import numpy as np
+from onnx import TensorProto
+from onnx.helper import make_node, make_tensor_value_info
+
+from ..helper_classes import BuildAModel
+from e2e_testing.registry import register_test
+from e2e_testing.storage import TestTensors
+
+class NonMaxSuppressionTestBasic(BuildAModel):
+    def construct_i_o_value_info(self):
+        self.input_vi = [
+            make_tensor_value_info("boxes", TensorProto.FLOAT, [1, 10, 4]),
+            make_tensor_value_info("scores", TensorProto.FLOAT, [1, 1, 10]),
+            make_tensor_value_info("max_output_boxes_per_class", TensorProto.INT64, [1]),
+            make_tensor_value_info("iou_threshold", TensorProto.FLOAT, [1]),
+            make_tensor_value_info("score_threshold", TensorProto.FLOAT, [1]),
+        ]
+        # Create ValueInfoProto object for selected_indices tensor
+        self.output_vi = [make_tensor_value_info("selected_indices", TensorProto.INT64, [1, 3])]
+
+    def construct_nodes(self):
+        node = make_node(
+            "NonMaxSuppression",
+            inputs=[
+                "boxes",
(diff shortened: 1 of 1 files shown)


lpy:
Looks like I can't add reviewers, @zjgarvey Could you please take a look? Thanks!

alt_e2eshark/onnx_tests/operators/non_max_suppression.py
```diff
@@ -0,0 +1,112 @@
...
+from onnx.helper import make_node, make_tensor_value_info
+
+from ..helper_classes import BuildAModel
+from e2e_testing.registry import register_test
+from e2e_testing.storage import TestTensors
+
+class NonMaxSuppressionTestBasic(BuildAModel):
+    def construct_i_o_value_info(self):
+        self.input_vi = [
+            make_tensor_value_info("boxes", TensorProto.FLOAT, [1, 10, 4]),
```
ScottTodd, line 17: FYI there are existing tests for this operator and similar inputs in both SHARK-TestSuite and iree-test-suites, both derived from the upstream tests at https://github.com/onnx/onnx/blob/main/onnx/backend/test/case/node/nonmaxsuppression.py
* upstream generated test case 1: https://github.com/onnx/onnx/tree/main/onnx/backend/test/data/node/test_nonmaxsuppression_identical_boxes
* upstream generated test case 2: https://github.com/onnx/onnx/tree/main/onnx/backend/test/data/node/test_nonmaxsuppression_center_point_box_format

Generated tests are defined in SHARK-TestSuite here:
* https://github.com/nod-ai/SHARK-TestSuite/blob/main/alt_e2eshark/onnx_tests/operators/generate_node.py

... and in iree-test-suites here:
* https://github.com/iree-org/iree-test-suites/blob/main/onnx_ops/onnx/node/generated/test_nonmaxsuppression_identical_boxes/model.mlir
* https://github.com/iree-org/iree-test-suites/blob/main/onnx_ops/onnx/node/generated/test_nonmaxsuppression_center_point_box_format/model.mlir

lpy: Thanks for pointing this out. Does that mean I should instead just update the expected failure list if llvm/torch-mlir#3976 is merged?

ScottTodd: I expect that when IREE pulls in your commit to torch-mlir, the CI workflow will "fail" (unexpectedly passing / XPASS) and then we'll remove these XFAILS:
* https://github.com/iree-org/iree/blob/4a7af8728126c722d4b293af1bc8bf21b732164a/tests/external/iree-test-suites/onnx_ops/onnx_ops_cpu_llvm_sync.json#L112
* https://github.com/iree-org/iree/blob/4a7af8728126c722d4b293af1bc8bf21b732164a/tests/external/iree-test-suites/onnx_ops/onnx_ops_gpu_rocm_rdna3.json#L107
* https://github.com/iree-org/iree/blob/4a7af8728126c722d4b293af1bc8bf21b732164a/tests/external/iree-test-suites/onnx_ops/onnx_ops_gpu_vulkan.json#L149

Similar to how https://github.com/iree-org/iree/pull/19659 fixed some `test_maxpool_2d_ceil_output_size_reduce_by_one` tests.

ScottTodd: Adding new hand-written / Python tests to `alt_e2eshark/onnx_tests/operators/` is okay... especially as a learning exercise to see how all the pieces fit together when working with a framework and the underlying tools directly. I personally would suggest against making such hand-written tests cover all cases exhaustively though, since that duplicates work that is already done upstream and in generated test suites and introduces more code to maintain.

PR #441, opened 2025-02-07
Update tokenizer input registry for HF models

vinayakdsci:
Updates tokenizer input registry for failing HuggingFace models with extra tokenizer attributes.


pdhirajkumarprasad (approved):
approve

PR #444, opened 2025-02-13
Allows duplicate finder tool to search within multiple run directories

zjgarvey:
If running 

```
python run.py -t <some_test_1> -v -r test-run1 --get-metadata
```
and 

```
python run.py -t <some_test_2> -v -r test-run2 --get-metadata
```
Then 

```
python utils/find_duplicate_models.py -r . 
```
from the `alt_e2eshark` directory will identify duplicates from both `./test-run1/` and `./test-run2/`.


pdhirajkumarprasad (approved):
LGTM

PR #447, opened 2025-02-24
Fixes native inference input size mistmatch issue

amd-vivekag:
Adds extra input during construct_input phase to avoid input vs session_input size mismatch issue during native inference.


pdhirajkumarprasad (approved):
LGTM

PR #451, opened 2025-02-27
Fixes opset_version issue, OOM issue

amd-vivekag:
- Fixes opset_version issue for following models:
  - hf_llama-7b
  - hf_oasst-sft-4-pythia-12b-epoch-3.5
  - hf_Qwen2.5-1.5B-Instruct
  - hf_Qwen2.5-7B-Instruct
  - hf_Qwen2-7B-Instruct
  - hf_TinyLlama-1.1B-Chat-v1.0
  - hf_vicuna-7b-v1.5

- Fixes OOM issue for hf_StableBeluga2
- removes hf_wasmai-7b-v1 model from the hf list


alt_e2eshark/e2e_testing/framework.py
```diff
@@ -64,6 +64,11 @@ def __init__(
         self.extra_options = ExtraOptions()
         self.update_extra_options()
 
+
+    def update_model_without_ext_data(self):
```
zjgarvey, line 68: Add a docstring with some info about what this does and why we have it. E.g.,

```python
"""For large models, which fail opset_version updating, use this method to update without loading external data."""
```

alt_e2eshark/onnx_tests/models/hf_models.py
```diff
@@ -235,10 +267,101 @@ def build_repo_to_model_map():
...
+                "attention_mask": {0: "B", 1: "L"},
+                "output": {0: "B", 1: "L"},
+            }
+        )
+        inputs = self.construct_inputs().data
+        print(f"{inputs=}", file=sys.stderr)
+
+        print("Exporting model to ONNX (this might take a while)...")
+        import torch
+        import torch.onnx
```
zjgarvey, line 343: I don't think we need this import?

alt_e2eshark/onnx_tests/models/hf_models.py
```diff
@@ -235,10 +267,101 @@ def build_repo_to_model_map():
...
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError(
+                "The package iree-turbine (try pip install iree-turbine) "
+                "is used to save hf model parameters to irpa format."
+            ) from e
+
+        print("Saving params (this might take a while)...")
+        self.param_path = os.path.join(self.cache_dir, self.name)
+        save_module_parameters(self.param_path, model)
+        print("Params saved.")
```
zjgarvey, line 329: You aren't using these params, so I wouldn't save them.

zjgarvey: Okay, so it looks like you aren't exporting params in `torch.onnx.export` below. 

Either:

1. don't use IREE turbine to save params at all AND `export_params=True` with `torch.onnx.export`
2. use IREE turbine to save params AND `export_params=False` `keep_initializers_to_inputs=True` `do_constant_folding=False` in `torch.onnx.export` AND change the `RuntimeOptions` to call both param files (the iree-turbine saved ones and the additional ones saved during mlir import) AND change the importer options to convert inputs to params.

zjgarvey (commented):
I think something is a bit wrong with the large model export path, but the rest looks good.

amd-vivekag:
> I think something is a bit wrong with the large model export path, but the rest looks good.

Can you please confirm if the following code is fine now?
```
    def export_large_models(self):
        from transformers import AutoModelForCausalLM
        print("\nLoading Hugging Face model...")
        model = AutoModelForCausalLM.from_pretrained(
            self.model_repo_path,
            cache_dir=self.cache_dir,
            #torch_dtype=self.torch_dtype,
        )
        print("Model loaded.")

        dynamic_axes = (
            {
                "input_ids": {0: "B", 1: "L"},
                "attention_mask": {0: "B", 1: "L"},
                "output": {0: "B", 1: "L"},
            }
        )
        inputs = self.construct_inputs().data

        print("Exporting model to ONNX (this might take a while)...")
        with torch.inference_mode():
            torch.onnx.export(
                model,
                (inputs[0], inputs[1]),
                self.model,
                export_params=True,
                do_constant_folding=True,
                keep_initializers_as_inputs=False,
                opset_version=19,
                dynamo=self.dynamo,
                input_names=["input_ids", "attention_mask"],
                output_names=["output"],
                dynamic_axes=dynamic_axes,
            )
        print("ONNX model exported.")

        if not os.path.isfile(self.model):
            raise RuntimeError(
                f"Torch onnx export failed to produce an onnx model at {self.model}"
            )

```

alt_e2eshark/e2e_testing/framework.py
amd-vivekag: Done

alt_e2eshark/onnx_tests/models/hf_models.py
amd-vivekag: removed

alt_e2eshark/onnx_tests/models/hf_models.py
amd-vivekag: Implemented first suggestion, by setting export_params=True and removed save params call. The other options were left as it is, please suggest if there is any particular option which should be set or unset.

zjgarvey (approved):

PR #452, opened 2025-02-27
Add (failing) test for bool tensor constants

rkayaith:


zjgarvey:
Ah, interesting. 

I figured out how to plumb through no-input tests, and I'll take a look at what is going on here.

zjgarvey:
It's happening because of:

<https://github.com/llvm/torch-mlir/blob/670618379e17719505a881b8f447602a0f4f5757/python/torch_mlir/extras/onnx_importer.py#L1135-L1139>

I'm not sure why we were using packbits there. Maybe we can try removing it and doing `signless=True`? I'm not sure.

rkayaith:
`packbits` is right, but we also need to specify the shape and MLIR type:
```
    onnx.TensorProto.DataType.BOOL: lambda tp: DenseElementsAttr.get(
        np.packbits(
            np.asarray(tp.int32_data, dtype=np.bool_).reshape(tp.dims),
            axis=None,
            bitorder="little",
        ),
        shape=tp.dims,
        type=IntegerType.get_signless(1)
    ),
```
that gets things through compilation, but now `iree-run-module` is failing 🤔

rkayaith:
closing since the bug here is covered by a lit test now, and I'm not sure there's value in an end to end test for this

